PR #340, opened 2020-04-23
Use a CUDAGuard when running Torch models

VivekPanyam:
This PR ensures that we're running on the correct device even if something else calls `cudaSetDevice` before running inference.

This fixes a class of issues where another piece of code changes the current device for the current thread. For example, this can happen if TF and Torch run together on the same threadpool. TF will call `cudaSetDevice` and cause torch to break if it runs on the same thread in the future.

This can cause some obscure cuDNN errors and generally hard-to-debug issues.

### DIFF
--- a/source/neuropod/backends/torchscript/torch_backend.cc
+++ b/source/neuropod/backends/torchscript/torch_backend.cc
@@ -7,6 +7,10 @@
 #include "neuropod/backends/torchscript/type_utils.hh"
 #include "neuropod/internal/tensor_types.hh"
 
+#ifndef __APPLE__
+#include <c10/cuda/CUDAGuard.h>
+#endif
+
 #include <caffe2/core/macros.h>
 
 #include <iostream>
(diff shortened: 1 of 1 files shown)


source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -291,6 +295,16 @@ std::unique_ptr<NeuropodValueMap> TorchNeuropodBackend::infer_internal(const Neu
 {
     torch::NoGradGuard guard;
 
+#ifndef __APPLE__
+    // Make sure we're running on the correct device
+    std::unique_ptr<at::cuda::CUDAGuard> device_guard;
```
vkuzmin-uber, line 300: #include \<memory\>

source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -291,6 +295,16 @@ std::unique_ptr<NeuropodValueMap> TorchNeuropodBackend::infer_internal(const Neu
...
+    const auto                           model_device = get_torch_device(DeviceType::GPU);
+    if (model_device.is_cuda())
+    {
+        device_guard = stdx::make_unique<at::cuda::CUDAGuard>(model_device);
```
vkuzmin-uber, line 304: I guess we can use std:: here not stdx:: because Neuropod became C++14 recently, right?

PR #356, opened 2020-05-21
[Python] Experimental support for specifying model dependencies

VivekPanyam:
### Summary:

This PR allows python/pytorch models to specify their python dependencies at packaging time. These dependencies are used to generate a `requirements.lock` file that is stored within the model.

When loading a python model, if a requirements.lock file is found, the dependencies are automatically installed into `~/.neuropod/pythonpackages/py{MAJOR_VERSION}{MINOR_VERSION}/` (if they don't already exist).

These python packages are installed in a way where we can independently load each python package we want.

### Test Plan:

Added a test that packages, loads, and runs inference with a python model that requires `scikit-learn`.

### DIFF
--- a/source/python/neuropod/backends/python/executor.py
+++ b/source/python/neuropod/backends/python/executor.py
@@ -25,6 +25,7 @@ import numpy as np
 
 from neuropod.backends.neuropod_executor import NeuropodExecutor
 from neuropod.utils.hash_utils import sha256sum
+from neuropod.utils.pip_utils import load_deps
 
 # Workaround for https://bugs.python.org/issue32573
 if not hasattr(sys, "argv"):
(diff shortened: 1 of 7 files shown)


source/python/neuropod/backends/python/executor.py
```diff
@@ -73,6 +74,15 @@ def __init__(self, neuropod_path, load_custom_ops=True):
         entrypoint_package_path = model_config["entrypoint_package"]
         entrypoint_fn_name = model_config["entrypoint"]
 
+        if is_native:
```
vkuzmin-uber, line 77: in case of is_native = False, do you expect that requirements.lock should be installed manually?

VivekPanyam: Added more details in the comment:

```
# For other code paths (e.g. when calling `load_neuropod` from python with `_always_use_native=False`),
# the user is responsible for ensuring that all dependencies are installed in the environment
```

PR #361, opened 2020-05-27
[WIP][Torch] Seeds / Determinism

### LINKED ISSUE #325: [Torch] Seeds / Determinism
Enable setting a random seed when running a model + set cuDNN to deterministic mode

Have a branch locally, need to put up a PR

VivekPanyam:
This PR implements Torch specific seed and determinism options.

Closes #325

### DIFF
--- a/source/neuropod/options.hh
+++ b/source/neuropod/options.hh
@@ -75,6 +75,23 @@ struct RuntimeOptions
 
     // Whether or not to disable shape and type checking when running inference
     bool disable_shape_and_type_checking = false;
+
+    // EXPERIMENTAL
+    // A seed to use when running a graph
+    // Note: this currently only applies to TorchScript models
+    int64_t seed = -1;
+
+    // EXPERIMENTAL
+    // Whether or not to run in deterministic mode. See https://pytorch.org/docs/stable/notes/randomness.html#cudnn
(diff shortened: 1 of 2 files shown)


selitvin (approved):

source/neuropod/options.hh
```diff
@@ -75,6 +75,23 @@ struct RuntimeOptions
...
+    // EXPERIMENTAL
+    // A seed to use when running a graph
+    // Note: this currently only applies to TorchScript models
+    int64_t seed = -1;
+
+    // EXPERIMENTAL
+    // Whether or not to run in deterministic mode. See https://pytorch.org/docs/stable/notes/randomness.html#cudnn
+    // Note: this currently only applies to TorchScript models and affects all torchscript models in the process.
+    // Should only be used with OPE to avoid this issue.
+    bool torch_cudnn_deterministic = false;
```
vkuzmin-uber, line 88: My personal perferences:

Even we know that this is for Torch only for now, I would call it: backends_cudnn_deterministic and backends_cudnn_benchmark. And then comment specifies that currently this is Torch only. So, we are going to add more options probably in future and it is better to keep them generic and name accordingly (if only 1 backend supports it - fine, but it helps to see gaps/difference between backends and even can be a start point to add options to other framework.

PR #390, opened 2020-06-16
Adding Ludwig to the documentation

w4nderlust:
### Summary:

This PR adds links to Ludwig in the documentation and instructions on how to use Neuropod with Ludwig.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -301,6 +301,7 @@ python -m ludwig.utils.neuropod_utils --ludwig_model_path <LUDWIG_OUTPUT_DIRECTO
 ```
 
 By default the package will be a ZIP file, but you can specify `--package_as_dir` if you prefer the output Neuropod package to be a directory.
+Also note that the Python version should be 3.7+.
 For more details check the [Ludwig User Guide](https://ludwig-ai.github.io/ludwig-docs/user_guide/#neuropod-export).
     
     
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -287,14 +287,14 @@ create_keras_neuropod(
 
 ### Ludwig
 
-Ludwig contains a utility to export model to the Neuropod format.
+Ludwig contains a utility to export a model as a neuropod.
 If we train a Ludwig model like this:
 
 ```
 ludwig train --data_csv <MY_DATA>PATH> --model_definition <MY_MODEL_DEFINITION> --output_directory <LUDWIG_OUTPUT_DIRECTORY>
 ```
 
-We can package it as a Neuropod by running:
+We can package it as a neuropod by running:
 
 ```
 python -m ludwig.utils.neuropod_utils --ludwig_model_path <LUDWIG_OUTPUT_DIRECTORY>/<MODEL_NAME>/model --neuropod_path <NEUROPOD_OUTPUT_PATH>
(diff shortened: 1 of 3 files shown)


docs/tutorial.md
```diff
@@ -285,6 +285,24 @@ create_keras_neuropod(
...
+If we train a Ludwig model like this:
+
+```
+ludwig train --data_csv <MY_DATA>PATH> --model_definition <MY_MODEL_DEFINITION> --output_directory <LUDWIG_OUTPUT_DIRECTORY>
+```
+
+We can package it as a Neuropod by running:
+
+```
+python -m ludwig.utils.neuropod_utils --ludwig_model_path <LUDWIG_OUTPUT_DIRECTORY>/<MODEL_NAME>/model --neuropod_path <NEUROPOD_OUTPUT_PATH>
```
vkuzmin-uber, line 300: Can you mention here if output (neuropod model) is going to be a ZIP file or dir by default. Is it controlled by a flag here?

vkuzmin-uber, line 300: I see you use "python", is it important for Ludwig model what python version is used, should it be consistent with Neuropod's python (there are different packages)? I personally think we should make it clear to avoid confusions of new users.

w4nderlust: it uses the default parameter, so it's True by default, but i can add a flag for it if you believe ti would be useful. In general any suggestion is more than welcome on improving this call: https://github.com/uber/ludwig/blob/master/ludwig/utils/neuropod_utils.py#L277

w4nderlust: Ludwig works for python 3.5+, but is tested only on 3.6+, although when working on the deployment on some models in Michelangelo we discovered a but in the embedded python interpreter for 3.6 that was solved in 3.7, so really i should specify to use python 3.7 (haven't tested 3.8 yet). Does that sound good?

docs/tutorial.md
```diff
@@ -285,6 +285,26 @@ create_keras_neuropod(
     `create_keras_neuropod` runs inference with the test data immediately after creating the neuropod. Raises a `ValueError` if the model output does not match the expected output.
 
 
+### Ludwig
+
+Ludwig contains a utility to export model to the Neuropod format.
```
VivekPanyam, line 290: nit: `Ludwig contains a utility to export a model as a neuropod.`

docs/tutorial.md
```diff
@@ -285,6 +285,26 @@ create_keras_neuropod(
...
+
+We can package it as a Neuropod by running:
+
+```
+python -m ludwig.utils.neuropod_utils --ludwig_model_path <LUDWIG_OUTPUT_DIRECTORY>/<MODEL_NAME>/model --neuropod_path <NEUROPOD_OUTPUT_PATH>
+```
+
+By default the package will be a ZIP file, but you can specify `--package_as_dir` if you prefer the output Neuropod package to be a directory.
+Also note that the Python version should be 3.7+.
+For more details check the [Ludwig User Guide](https://ludwig-ai.github.io/ludwig-docs/user_guide/#neuropod-export).
```
VivekPanyam, line 305: It may also be worth mentioning that your runtime environment requires having all of Ludwig's deps installed.

This isn't ideal and am currently working on a fix (see #356), but if we're going to add Ludwig to the docs now, we should probably explain. Maybe add something like

```
!!! note
    Currently, running a Ludwig model requires having all of [Ludwig's dependencies](https://github.com/uber/ludwig/blob/master/requirements.txt) installed in your runtime environment. This will be improved in a future release.
```

Issue #374, opened 2020-06-02
AttributeError: module 'sys' has no attribute 'argv'


irasit:
Got this error when loading a model that contains sys.argv
```
LoadNeuropod modelPath = user_model/neuropod.zip failed: AttributeError: module 'sys' has no attribute 'argv'
At:
  /[redacted]/env/lib/python3.6/site-packages/ludwig/contrib.py(35): contrib_import
  /[redacted]/env/lib/python3.6/site-packages/ludwig/api.py(38): <module>
  <frozen importlib._bootstrap>(219): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(678): exec_module
  <frozen importlib._bootstrap>(665): _load_unlocked
  <frozen importlib._bootstrap>(955): _find_and_load_unlocked
  <frozen importlib._bootstrap>(971): _find_and_load
  /tmp/tmp17kynda9.neuropod_python_symlinks/aa723e1c_34fd_4df9_b1de_df9076e88879/ludwig/neuropod_export.py(10): <module>
  <frozen importlib._bootstrap>(219): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(678): exec_module
  <frozen importlib._bootstrap>(665): _load_unlocked
  <frozen importlib._bootstrap>(955): _find_and_load_unlocked
  <frozen importlib._bootstrap>(971): _find_and_load
  <frozen importlib._bootstrap>(994): _gcd_import
  /[redacted]/env/lib/python3.6/importlib/__init__.py(126): import_module
  /[redacted]/env/lib/python3.6/site-packages/neuropod-0.1.0-py3.6.egg/neuropod/backends/python/executor.py(120): __init__
  /[redacted]/env/lib/python3.6/site-packages/neuropod-0.1.0-py3.6.egg/neuropod/loader.py(29): load_neuropod
```

This issue related to this [python issue](https://bugs.python.org/issue32573) , it only happen in embedded python system, and fixed in py3.8.

But to make it backward compatible, we can add this work around 
```
import sys
if not hasattr(sys, 'argv'):
    sys.argv  = [''] 
```

irasit:
After adding the fix in load_model, this problem still happened in another place. 
[...]

PR #393, opened 2020-06-23
Move the embedded python interpreter sys.argv fix into the python backend

irasit:
### Summary
Move the `sys.argv` fix for embedded python interpreters (originally added in #375) from `loader.py` to `executor.py` in the python backend.

This is necessary because the C++ PythonBridge uses the python executor directly and bypasses `loader.py`.

More details in #374

### Test Plan
Added a new test that confirms `sys.argv` is available from within python models.

_Note: This description was edited by @VivekPanyam to add details and improve clarity_

### DIFF
--- a/source/python/neuropod/backends/python/executor.py
+++ b/source/python/neuropod/backends/python/executor.py
@@ -27,9 +27,7 @@ from neuropod.backends.neuropod_executor import NeuropodExecutor
 from neuropod.utils.hash_utils import sha256sum
 
 # Workaround for https://bugs.python.org/issue32573
-if not hasattr(sys, "argv") and (
-    sys.version_info.major == 3 and sys.version_info.minor < 8
-):
+if not hasattr(sys, "argv"):
     sys.argv = [""]
 
 # Create the neuropod package symlink directory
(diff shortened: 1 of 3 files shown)


irasit:
> Can you remove the code you added to `loader.py` in your previous PR?
> 
> https://github.com/uber/neuropod/blob/4f5764af8f29d1ca241f9576cf4ea5bae5c025dd/source/python/neuropod/loader.py#L23-L27
> 
> We shouldn't include that code in both the loader and the python backend.
> 
> Also, can you update the PR description to match the default format (`Summary` and `Test Plan`)?

Removed the fix from loader, and updated the comment.

irasit:
> Also it looks like lint failed (see https://neuropod.ai/developing/#lint-and-static-analysis).
> 
> (You'll need to sign the CLA as well - see the first comment on this PR)

Test case is added. Also remove the version check so it will also fix py2.

irasit:
> Looks good! Can you fix lint and sign the CLA (see my comment above: [#393 (review)](https://github.com/uber/neuropod/pull/393#pullrequestreview-436132914))?

Fixed the lint.

VivekPanyam:
_Note: Before merging, I edited the PR title, description, and test plan to be more detailed_

cc @irasit

PR #400, opened 2020-07-17
[C API] Initial implementation

VivekPanyam:
### Summary:

An initial basic implementation of a C API for Neuropod.

### Test Plan:

Added a test that loads a model, runs inference, and verifies the output. More unit tests and integration tests should be added in the future

### DIFF
--- a/source/neuropod/bindings/c/c_api_internal.h
+++ b/source/neuropod/bindings/c/c_api_internal.h
@@ -1,0 +1,25 @@
+/* Copyright (c) 2020 UATC, LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


source/neuropod/bindings/c/c_api_internal.h
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+#pragma once
+
+#include "neuropod/neuropod.hh"
+
+struct NP_Neuropod
+{
+    std::unique_ptr<neuropod::Neuropod> model;
```
vkuzmin-uber, line 22: nit: if unique_ptr is used explicitly, it makes sense to include <memory> explicitly too. The same about shared_ptr

source/neuropod/bindings/c/np_tensor_allocator_internal.h
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) 2020 UATC, LLC
```
vkuzmin-uber, line 1: btw, what is UATC, is it Uber ATG abbrevation?

source/neuropod/bindings/c/c_api_internal.h
VivekPanyam: Fixed

source/neuropod/bindings/c/np_tensor_allocator_internal.h
VivekPanyam: `UATC, LLC == Uber ATG`

PR #401, opened 2020-07-17
[Java API] Neuropod Java API Interface Proposal

weijiadeng-uber:
### Summary:
The Java API Interface for Neuropod Java API.

The NeuropodNativeTensor class will be implemented as the following: It can be created by both C++ side(as inference result) and Java side. It has a private field buffer. For the objects created by C++ side, the buffer is wrapped from the raw memory pointer from the C++ side tensor. For the objects created by Java side, the buffer is a direct native bytebuffer from the Java side. The C++ tensor is created from the Java bytebuffer's memory. User won't tell the difference between these two types of NeuropodNativeTensor, unless they try to convert it to a pure Java tensor NeuropodJavaTensor. NeuropodNativeTensor created by the C++ side will have to trigger a copy to make the memory section is allocated by the Java side.

### DIFF
--- a/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/Neuropod.java
+++ b/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/Neuropod.java
@@ -1,0 +1,104 @@
+package com.uber.neuropod;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This class holds the information of a Neuropod model. It has its underlying C++ Neuropod
+ * object, should call close() function to free the C++ side object when finish using the Neuopod object
+ */
+public class Neuropod extends NativeClass {
+
+    /**
+     * Load a model from the file path and use default options
+     *
+     * @param neuropodPath the neuropod model path
+     */
+    public Neuropod(String neuropodPath) {}
+
+    /**
+     * Load a model from the file path and use provided runtime options
+     *
+     * @param neuropodPath the neuropod model path
+     * @param options      the runtime options
+     */
+    public Neuropod(String neuropodPath, RuntimeOptions options) { }
+
+    /**
+     * Gets the model name.
+     *
+     * @return the name
+     */
+    public String getName() {
+        return null;
+    }
+
+    /**
+     * Gets the platform name.
+     *
+     * @return the platform
+     */
+    public String getPlatform() {
+        return null;
+    }
+
+
+    /**
+     * Perform the inference calculation on the given input data
+     *
+     * @param inputs the inputs
+     * @return the inference result
+     */
+    public Map<String, NeuropodTensor> infer(Map<String,NeuropodTensor> inputs) {
+        return null;
+    }
+
+    /**
+     * Perform the inference calculation on the given input data, only get output tensor
+     * with the keys specified by requestOutputs.
+     *
+     * @param inputs         the inputs
+     * @param requestedOutputs only tensor with these keys will be output
+     * @return the inference result
+     */
+    public Map<String, NeuropodTensor> infer(Map<String, NeuropodTensor> inputs, List<String> requestedOutputs) {
+        return null;
+    }
+
+    /**
+     * Gets input tensor specs
+     *
+     * @return the input specs
+     */
+    public List<TensorSpec> getInputs() {
+        return null;
+    }
+
+    /**
+     * Gets output tensor specs
+     *
+     * @return the output specs
+     */
+    public List<TensorSpec> getOutputs() {
+        return null;
+    }
+
+    /**
+     * Load model. Used when setLoadModelAtConstruction is set to false in RuntimeOptions
+     */
+    public void loadModel() {}
+
+    /**
+     * Gets a tensor allocator based on the backend of the model.
+     *
+     * @return the tensor allocator
+     */
+    public NeuropodTensorAllocator getTensorAllocator() {return null;}
+
+    /**
+     * Gets a generic tensor allocator.
+     *
+     * @return the generic tensor allocator
+     */
+    public static NeuropodTensorAllocator getGenericTensorAllocator() {return null;}
+}
--- a/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
+++ b/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
@@ -1,0 +1,146 @@
+package com.uber.neuropod;
+
+import java.io.Serializable;
+import java.nio.*;
+import java.util.List;
+
+/**
+ * This class is used for holding tensor data. It has its underlying C++ NeuropodTensor
+ * object, should call close() function to free C++ side object when finish using the
+ * NeuropodTensor object.
+ * This object can be created by the Java side or by the C++ side(as an inference result).
+ */
+public class NeuropodTensor extends NativeClass implements Serializable {
+     /**
+      * Get the dims array which represnents the shape of a tensor.
+      *
+      * @return the shape array
+      */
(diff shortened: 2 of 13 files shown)


source/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -0,0 +1,131 @@
...
+      * @return the list
+      */
+     List<String> toStringList();
+
+     /**
+      * Gets an int element at the given index.
+      *
+      * Can only be used when the tensor is INT32_TENSOR.
+      *
+      * @param index the index
```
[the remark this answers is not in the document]
weijiadeng-uber: The input is an index array. For example, for a 3rd-order tensors, to get certain element the user should call tensor.getInt(x, y, z) to get the element at index (x, y, z)

source/java/src/main/java/com/uber/neuropod/TensorType.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    static {
+        for (TensorType tensorType : TensorType.values()) {
+            map.put(tensorType.value, tensorType);
+        }
+    }
+
+    public static TensorType valueOf(int dataType) {
+        return (TensorType) map.get(dataType);
```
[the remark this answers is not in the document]
weijiadeng-uber: The enum to int and int to enum conversion is for internal usage only, I'll make them protected.

source/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -0,0 +1,131 @@
+package com.uber.neuropod;
+
+import java.util.List;
+
+/**
+ * The interface Neuropod tensor, which stores the data of a tensor. NeuropodTensors are backed
+ * by a buffer, and the buffer may be allocated by the C++ side(inference result) or the Java side.
+ */
+interface NeuropodTensor {
```
VivekPanyam, line 9: As I mentioned above, I think we should combine `NeuropodJavaTensor` and `NeuropodNativeTensor` and turn this interface into a class.

source/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -0,0 +1,131 @@
...
+
+     /**
+      * Flatten the tensor data and convert it to a double array.
+      *
+      * Can only be used when the tensor is DOUBLE_TENSOR. Will trigger
+      * a copy if the backed buffer is not a DoubleBuffer backed by an array.
+      *
+      * @return the double [ ]
+      */
+     double[] toDoubleArray();
```
VivekPanyam, line 69: Doesn't match the RFC (and as we talked about this won't work without copies for direct ByteBuffers). We should replace these with methods that return `IntBuffer`, `LongBuffer`, `FloatBuffer`, and `DoubleBuffer`.

Ideally we wouldn't have all of these methods on `NeuropodTensor` because most of them won't be valid on any given tensor. For now, it should be fine, but we should think about other approaches in the future. In C++, we have a templated subclass of `NeuropodTensor` called `TypedNeuropodTensor<T>` where T can be any of the supported tensor types.

Not sure how easily we can do something similar with generics in Java since we can't do things like specializations. Maybe a more straightforward approach would be to create a subclass for each type (e.g. `FloatNeuropodTensor`, `StringNeuropodTensor`, etc.).

source/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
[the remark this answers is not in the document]
VivekPanyam: Yeah add some docs, but this looks good to me

source/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
weijiadeng-uber: The problem is generic types for primitive types is not supported in Java, so having different functions for different types seems to be a common solution(although for certain tensor only one function may apply). See the tensorflow Java API for more details:[](https://www.tensorflow.org/api_docs/java/reference/org/tensorflow/Tensor#writeTo(java.nio.LongBuffer))

source/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -0,0 +1,11 @@
+package com.uber.neuropod;
```
VivekPanyam, line 1: I think you forgot to remove this file (there's already one under `source/neuropod/bindings/java`)

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -0,0 +1,146 @@
...
+      */
+     public TensorType getTensorType() {return null;}
+
+     /**
+      * Flatten the tensor data and convert it to a long buffer.
+      * <p>
+      * Can only be used when the tensor is INT64_TENSOR. Will trigger
+      * a copy.
+      *
+      * @return an array backed LongBuffer
```
VivekPanyam, line 41: Do we need these to be array backed buffers?

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensorAllocator.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+
+/**
+ * The factory type for NeuropodTensor, can be obtained from a neuropod model or as a generic allocator.
+ * Should call close method when finish using it.
+ */
+public class NeuropodTensorAllocator extends NativeClass {
+    /**
+     * Create a NeuropodTensor based on given ByteBuffer, dims array and tensorType by
+     * the allocator. The created tensor will have the input tensorType
```
VivekPanyam, line 13: nit: Remove `by the allocator` (similarly for the below methods)

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/Neuropod.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Perform the inference calculation on the given input data, only get output tensor
+     * with the keys specified by requestOutputs.
+     *
+     * @param inputs         the inputs
+     * @param requestOutputs only tensor with these keys will be output
+     * @return the inference result
+     */
+    public Map<String, NeuropodTensor> infer(Map<String, NeuropodTensor> inputs, List<String> requestOutputs) {
```
VivekPanyam, line 64: nit: `requestedOutputs`

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
weijiadeng-uber: Yeah, this way users can know they can call array() function to get the array form of the data.

VivekPanyam: A user can just call [hasArray](https://docs.oracle.com/javase/7/docs/api/java/nio/LongBuffer.html#hasArray()) to see if they can call `array()`.

I don't think we want to commit to always copying the data just because a user could call `array()`. Since we provide elementwise access (below), I think it's okay for us to return non array-backed buffers here.

Users can also do this copy on their end so they're only paying the performance penalty of the features they're using.

weijiadeng-uber: I see. I've deleted the an array backed part.

PR #402, opened 2020-07-24
[Java API] Add build configuration and library loading

weijiadeng-uber:
### Summary:
Add bazel build file and methods to load native library.

### Test Plan:
```
bazel test //neuropod/bindings/java:*
```


source/neuropod/bindings/java/BUILD
```diff
@@ -0,0 +1,136 @@
...
+        "//neuropod:__subpackages__",
+    ],
+)
+
+java_binary(
+    name = "neuropod_java_fat_jar",
+    srcs = [":java_source"],
+    # TODO: Find a way to make these native libraries locate at /com/uber/neuropod/native/{OS}/{PLATFORM}
+    # in the jar file, instead of locating at the class root path.
+    classpath_resources = [":native_libraries"],
```
vkuzmin-uber, line 68: Should it be runtime_deps instead https://docs.bazel.build/versions/master/be/java.html#java_binary.runtime_deps ?

source/neuropod/bindings/java/README.md
```diff
@@ -1,4 +1,13 @@
 # The java api interface:
+0. To build a fat jar:
+```
+bazel build //neuropod/bindings/java:neuropod_java_fat_jar
```
vkuzmin-uber, line 4: Can you leave instruction for non-fat (thin) jar too?

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -1,19 +1,143 @@
...
+                                                                "libcaffe2_observers.so",
+                                                                "libfbjni.so",
+                                                                "libgomp-7c85b1e2.so.1",
+                                                                "libpytorch_jni.so",
+                                                                "libshm.so",
+                                                                "libtorch_cpu.so",
+                                                                "libtorch_global_deps.so",
+                                                                "libtorch_python.so",
+                                                                "libtorch.so",
+                                                                // TorchScript libraries darwin name
```
vkuzmin-uber, line 46: Not sure if this is critical, can it be platform specific, using compile time macro?

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -1,19 +1,143 @@
...
+                                                                "libc10.dylib",
+                                                                "libcaffe2.dylib",
+                                                                "libmklml.dylib",
+                                                                "libtorch.1.dylib",
+                                                                "libtorch.dylib",
+                                                                "libiomp5.dylib"
+                                                                );
+    private static final List<String> BIN_NAMES = Arrays.asList("neuropod_multiprocess_worker");
+    private static final String JNI_NAME = "neuropod_jni";
```
vkuzmin-uber, line 55: Isn't it libneuropod_jni or it doesn't need "lib" prefix here? If so, please leave comment to make it clear.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -1,19 +1,143 @@
...
+    private static final String TEMP_DIRECTORY_NAME = "neuropod_native_libraries";
+
     /**
      * Load jni library.
      */
     public static void load() {
+        if (isLoaded() || loadSharedLibrary() || loadEmbeddedLibrary()) {
+            return;
+        }
+        throw new UnsatisfiedLinkError("Load neuropod jni library failed");
```
vkuzmin-uber, line 67: maybe specify exact name libneuropo_jni?

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -1,19 +1,143 @@
...
+    private static final String JNI_NAME = "neuropod_jni";
+    private static final String JNI_EMBEDDED_NAME = "libneuropod_jni.so";
+    private static final int BUFFER_SIZE = 1<<20;
+    private static final String TEMP_DIRECTORY_NAME = "neuropod_native_libraries";
+
     /**
      * Load jni library.
      */
     public static void load() {
+        if (isLoaded() || loadSharedLibrary() || loadEmbeddedLibrary()) {
```
vkuzmin-uber, line 64: Interesting, is there guarantee that this is called w/o concurrency and so no multiple therads? Otherwise it will need some locking/memory barrier to guarantee correctness. (2 callers may get isLoaded() == false at the same time and proceed with calling load* methods)

If this is guaranteed, please specify in the comment of method that there is a guarantee on it.

source/neuropod/bindings/java/src/main/native/utils.cc
```diff
@@ -0,0 +1,27 @@
+#include "utils.h"
+
+#include "neuropod/neuropod.hh"
+
+#include <iostream>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <vector>
```
vkuzmin-uber, line 9: It includes a lot of unused headers. In C++ best practice is to include only explicitly used headers, I see std:string and jni.h only

source/neuropod/bindings/java/src/main/native/utils.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "neuropod/neuropod.hh"
+
+#include <string>
+#include <vector>
```
vkuzmin-uber, line 6: check what headers you really need

source/neuropod/bindings/java/BUILD
weijiadeng-uber: I use classpath_resources here because I want to test whether the Library loader can use the native libraries packed in the jar file. I think the fat jar will be used more widely because user does not need to install any external dependencies outside the java environment.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
weijiadeng-uber: Yeah, it does not need lib prefix here because when call System.loadLibrary java will auto complete the lib and so part based on the OS.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
weijiadeng-uber: We can have two list, one for linux and one for darwin.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
weijiadeng-uber: Yes, Java static initializers are thread safe

source/neuropod/bindings/java/src/main/native/utils.cc
```diff
@@ -0,0 +1,23 @@
+#include "utils.h"
+
+#include "neuropod/neuropod.hh"
```
vkuzmin-uber, line 3: Are util.h and neuropod.hh headers used here?

source/neuropod/bindings/java/src/main/native/utils.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "neuropod/neuropod.hh"
```
vkuzmin-uber, line 3: It is not used as far as I can see

source/neuropod/bindings/java/Makefile
```diff
@@ -0,0 +1,9 @@
+export MY_JNI_HEADERS=${PWD}/src/main/native
```
VivekPanyam, line 1: We shouldn't have makefiles anywhere.

Can we just generate the headers and check them in?

Since the implementation isn't autogenerated, I think generating the headers during the build might cause more problems than it's worth

source/neuropod/bindings/java/README.md
```diff
@@ -1,9 +1,19 @@
 # The java api interface:
-1. To load a model:
+0. Generate C++ headers:
+```
+make genHeaders
```
VivekPanyam, line 4: Remove this step - just check the headers in

.gitignore
```diff
@@ -16,3 +16,4 @@ bazel-testlogs
 source/python/neuropod/libneuropod.so
 source/python/neuropod/neuropod_multiprocess_worker
 source/python/neuropod/neuropod_native.so
+source/neuropod/bindings/java/src/main/native/com_uber_neuropod*.h
```
VivekPanyam, line 19: As in my other comments - let's just include the headers in the checked-in code

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -1,19 +1,129 @@
...
+        } catch (UnsatisfiedLinkError e) {
+            LOGGER.log(Level.WARNING, "Try loading shared library failed, should be fine if using the " +
+                    "library from a jar file: " + e.getMessage());
+            return false;
+        }
+    }
+
+    // Load the jni library from the jar file
+    private static boolean loadEmbeddedLibrary() {
+        // TODO: Should be /com/uber/neuropod/native/{OS}/{PLATFORM} once we figure out how to copy native libraries
```
VivekPanyam, line 56: nit: `TODO(weijiad):` and remove the `by weijiad` at the end

source/neuropod/bindings/java/BUILD
```diff
@@ -0,0 +1,170 @@
...
+        "@bazel_tools//src/conditions:darwin": ["-Wl,-rpath,@loader_path"],
+    }),
+    linkshared = True,
+    visibility = [
+        "//visibility:public",
+    ],
+    deps = [
+        ":jni_core",
+    ],
+)
```
VivekPanyam, line 159: This approach isn't going to work on linux (because it's going to attempt to build something named `libneuropod_jni.dylib` and fail (see the output of the failing CI for this PR).

Another solution that might work is to just use a genrule for the dylib (since we just care about changing the name). Maybe something like:

```
genrule(
    name = "copy_to_dylib",
    srcs = [":libneuropod_jni.so"],
    outs = ["libneuropod_jni.dylib"],
    cmd = "cp -f $< $@",
)
```

This isn't ideal, but we can improve it after we get a basic first version running. Ideally, we'd be able to provide the full name to `System.loadLibrary` (e.g. `libneuropod_jni.so`) and not have to do this during the build

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/LibraryLoader.java
```diff
@@ -1,19 +1,129 @@
...
+    }
+
+
+    private static boolean isLoaded() {
+        try {
+            return nativeIsLoaded();
+        } catch (UnsatisfiedLinkError e) {
+            return false;
+        }
     }
```
VivekPanyam, line 124: This feels a little hacky, but I think it's fine for now

source/neuropod/bindings/java/BUILD
```diff
@@ -0,0 +1,155 @@
...
+        "//conditions:default": ["external/local_jdk/include/linux"],
+    }),
+    visibility = [
+        "//neuropod:__subpackages__",
+    ],
+    deps = [
+        ":copy_jni_hdr_lib",
+        "//neuropod:neuropod_hdrs",
+    ],
+)
```
VivekPanyam, line 124: Can this be combined into the `libneuropod_jni.so` target below?

PR #405, opened 2020-08-02
[Java_API] Initial implementation of model loading

weijiadeng-uber:
### Summary:
The load model part of Neuropod JAVA API.
Currently we have to set useOpe to true to load a model.

### Test Plan:
```
./build/docker_build.sh
```


source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -0,0 +1,232 @@
...
+                                                                                        jlong   optHandle)
+{
+    auto opts = reinterpret_cast<neuropod::RuntimeOptions *>(optHandle);
+    try
+    {
+        auto                convertedPath = toString(env, path);
+        neuropod::Neuropod *ret           = nullptr;
+        if (isTestMode)
+        {
+            ret = new neuropod::Neuropod(convertedPath, detail::ope_backend_location_overrides, *opts);
```
weijiadeng-uber, line 74: The override is passed here

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+public class NeuropodTest {
+    private Neuropod model;
+    private static final String MODEL_PATH = "neuropod/tests/test_data/tf_addition_model/";
+
+    @Before
+    public void setUp() throws Exception {
+        LibraryLoader.load();
+        // Set the test mode to true to use override library path
+        LibraryLoader.setTestMode(true);
```
weijiadeng-uber, line 36: The test mode is set to true here.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_LibraryLoader.cc
```diff
@@ -31,3 +32,8 @@ JNIEXPORT void JNICALL Java_com_uber_neuropod_LibraryLoader_nativeExport(JNIEnv
     std::string oriPath = getenv("PATH");
     setenv("PATH", (oriPath + ":" + neuropod::jni::toString(env, libPath)).c_str(), 1 /* Overwrite */);
 }
+
+JNIEXPORT void JNICALL Java_com_uber_neuropod_LibraryLoader_nativeSetTestMode(JNIEnv *, jclass, jboolean mode)
+{
+    isTestMode = mode == JNI_TRUE;
```
weijiadeng-uber, line 38: This is how the test mode is set to true

weijiadeng-uber: ChangeLog: Add test for TensorSpec symbol case, fix the strange line counting on NeuropodTensor.java, add try catch block for delete functions, merge two nativeNew method for Neuropod into one.

source/neuropod/bindings/java/src/main/native/jclass_register.h
```diff
@@ -0,0 +1,37 @@
...
+
+extern jclass    com_uber_neuropod_Dimension;
+extern jmethodID com_uber_neuropod_Dimension_value_;
+extern jmethodID com_uber_neuropod_Dimension_symbol_;
+
+extern jclass com_uber_neuropod_TensorType;
+
+extern jclass com_uber_neuropod_NeuropodJNIException;
+
+extern bool isTestMode;
```
VivekPanyam, line 37: Can we wrap this in a `neuropod::jni` namespace / does it make sense to do so?

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -0,0 +1,215 @@
...
+        auto                convertedPath = toString(env, path);
+        neuropod::Neuropod *ret           = nullptr;
+        if (isTestMode)
+        {
+            ret = new neuropod::Neuropod(convertedPath, detail::ope_backend_location_overrides, opts);
+        }
+        else
+        {
+            ret = new neuropod::Neuropod(convertedPath, opts);
+        }
```
VivekPanyam, line 57: Is there a cleaner way we can do this? Can we expose the `default_backend_overrides` argument to java?

`isTestMode` and `setTestMode` seem like a workaround

weijiadeng-uber: ope_backend_location_overrides depends on current working path. I am afraid that the current working path in Java and C++ may be different in some cases.

source/neuropod/bindings/java/src/main/native/jclass_register.h
weijiadeng-uber: This is to improve the speed finding java class ID and method ID (https://stackoverflow.com/questions/10617735/in-jni-how-do-i-cache-the-class-methodid-and-fieldids-per-ibms-performance-r)

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
vkuzmin-uber: I also think it would be good to refactor code to not have isTestMode explicitly. And instead allow to pass/set ope_backend_location_overrides somehow. But at the same time I am ok if we proceed this way and get back to it later when get stable initial release.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_NeuropodTensorAllocator.cc
```diff
@@ -0,0 +1,35 @@
...
+
+#include <jni.h>
+
+JNIEXPORT void JNICALL Java_com_uber_neuropod_NeuropodTensorAllocator_nativeDelete(JNIEnv *env, jobject, jlong handle)
+{
+    try
+    {
+        auto allocator = reinterpret_cast<std::shared_ptr<neuropod::NeuropodTensorAllocator> *>(handle);
+        (*allocator).reset();
+        delete allocator;
```
vkuzmin-uber, line 29: delete 0 or delete nullptr is allowed and safe. But not *nullptr. I guess it should make sure we don't dereference nullptr here:

```
if (allocator != nullptr) {
        allocator->reset();
        delete allocator;
}
```

weijiadeng-uber: In the java side close method, we have if (nativeHandle == 0) { return;} to ensure the handle is non zero.

weijiadeng-uber:
Split the commit. The tensor allocator, get input spec and get output spec functions are moved to the next PR.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NativeClass.java
```diff
@@ -15,18 +15,72 @@
...
     @Override
-    public void close() throws Exception {
-        // Wrap the nativeDelete to close method so that we can do some common post-process here if needed.
+    public final void close() throws NeuropodJNIException {
+        // Wrap the nativeDelete to close method so that the IDE will have a warning
+        // if the object is not deleted, and will be auto deleted in a try catch block.
+        if (nativeHandle == 0) {
+            return;
+        }
+        this.nativeDelete(this.getNativeHandle());
```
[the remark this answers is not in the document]
weijiadeng-uber: We do not need the protection in this method because we will first check whether the nativeHandle is 0, if is zero nothing happens.

source/neuropod/bindings/java/src/main/native/utils.h
```diff
@@ -27,5 +29,12 @@ namespace jni
 // Convert jstring to cpp string
 std::string toString(JNIEnv *env, jstring target);
 
+// A wrapper for env->FindClass, will throw a cpp exception if the find fails.
+jclass findClass(JNIEnv *env, const char *name);
+
+// Throw a Java NeuropodJNIException exception. This throw is only valid for the Java side, in the cpp side it is just
+// like a normal function: it won't interupt the normal program work flow.
```
vkuzmin-uber, line 36: interRupt

vkuzmin-uber: I dont follow, is it going to be used as a "normal" function at all?

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_NeuropodTensorAllocator.cc
weijiadeng-uber: nativeDelete method is only called by close method, so this check always applies.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -0,0 +1,117 @@
...
+        return reinterpret_cast<jlong>(ret);
+    }
+    catch (const std::exception &e)
+    {
+        throwJavaException(env, e.what());
```
vkuzmin-uber, line 60: I don't follow when it will throw or when this is just a call, like a normal. Can you add a comment in a call where this is not a "throw" ?

source/neuropod/bindings/java/src/main/native/utils.h
weijiadeng-uber: It is not a throw in the C++ side function, instead it throws an exception in the Java side. When the program counter go back to the java side, the java will know there is an exception thrown.

For example, in the C++ side:

std::vector<int>* a = new std::vector<int>();
delete a;
throwJavaException(env, e.what());
delete a;

This will cause a segfault, because the throwJavaException won't interrupt the C++ side code, both delete is executed.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
weijiadeng-uber: For example, in the C++ side:

std::vector* a = new std::vector();
delete a;
throwJavaException(env, e.what());
delete a;

This will cause a segfault, because the throwJavaException won't interrupt the C++ side code, both delete is executed.

But for this code:

void jni_foo(env) {
         throwJavaException(JNIEnv *env, "foo");
}

And in the java side:
void main() {
       jni_foo();
}

This program will throw a NeuropodJNIException.

weijiadeng-uber: Or we can say it throws a Java NeuropodJNIException exception after the JNI call have finished.

Issue #407, opened 2020-08-06
Add a C API to neuropod


vkuzmin-uber:
## Feature
C API

We don't expect that somebody is using C API directly. This will be used by future CGo API mostly.

Connected to https://github.com/uber/neuropod/issues/294

## Describe the solution you'd like
Design is inspired by the TensorFlow C API and Uber CGO package that includes Bridge C/C++ bridge. Namely:
Internal C Headers
Public C Headers
C/C++ implementation of public and internal headers.

VivekPanyam:
> We don't expect that somebody is using C API directly. This will be used by future CGo API mostly.

I don't think this is accurate. We should build this API in a way where it provides a robust easy-to-use interface to use Neuropod from C (independent of the Go API)

vkuzmin-uber:
> > We don't expect that somebody is using C API directly. This will be used by future CGo API mostly.
> 
> I don't think this is accurate. We should build this API in a way where it provides a robust easy-to-use interface to use Neuropod from C (independent of the Go API)

Agree, just wanted to highlight that CGO is going to be a first user really, not only.

PR #409, opened 2020-08-06
[C API] Continue implementation

vkuzmin-uber:
### Summary:

Continue with basic implementation of a C API for Neuropod. Issue #407 

### Test Plan:

Added a test that loads a model and fails, runs inference successfully and fails, and verifies the name and platform.

### DIFF
--- a/source/neuropod/bindings/c/README.md
+++ b/source/neuropod/bindings/c/README.md
@@ -1,7 +1,13 @@
-# The C api interface:
+# C API
 
-1. To build & to unit test:
+***NOTE: This API is currently experimental***
+
+## Build
 ```
 bazel build //neuropod/bindings/c:c_api
+```
+
+## Run C unit tests
+```
 bazel test --test_output=errors //neuropod/tests:test_c_api
 ```
--- a/source/neuropod/bindings/c/README.md
+++ b/source/neuropod/bindings/c/README.md
@@ -1,7 +1,7 @@
 # The C api interface:
 
-1. To build &to unit test :
-``` 
-bazel build // neuropod/bindings/c:c_api
-bazel test --test_output=errors // neuropod/tests:test_c_api
+1. To build & to unit test:
+```
+bazel build //neuropod/bindings/c:c_api
+bazel test --test_output=errors //neuropod/tests:test_c_api
 ```
--- a/source/neuropod/bindings/c/README.md
+++ b/source/neuropod/bindings/c/README.md
@@ -1,0 +1,7 @@
+# The C api interface:
+
+1. To build &to unit test :
+``` 
+bazel build // neuropod/bindings/c:c_api
+bazel test --test_output=errors // neuropod/tests:test_c_api
+```
--- a/source/neuropod/bindings/c/c_api.cc
+++ b/source/neuropod/bindings/c/c_api.cc
@@ -35,18 +35,12 @@ void NP_LoadNeuropodWithOpts(const char *             neuropod_path,
     {
         *model          = new NP_Neuropod();
         (*model)->model = std::make_unique<neuropod::Neuropod>(neuropod_path);
-        if (status != nullptr)
-        {
-            NP_ClearStatus(status);
-        }
+        NP_ClearStatus(status);
     }
     catch (std::exception &e)
     {
-        if (status != nullptr)
-        {
-            status->code    = NEUROPOD_ERROR;
-            status->message = e.what();
-        }
+        status->code    = NEUROPOD_ERROR;
+        status->message = e.what();
     }
 }
 
@@ -96,25 +90,24 @@ void NP_InferWithRequestedOutputs(NP_Neuropod *              model,
     try
     {
         // By default empty collection of requested_ouputs expected.
+        // Note that it copies C-strings (0-terminated) from given array.
+        // User must guarantee that noutputs and requested_outputs address valid data..
         std::vector<std::string> rout;
-        for (size_t i = 0; i < noutputs; ++i)
+        if (requested_outputs != nullptr)
         {
-            rout.push_back(requested_outputs[i]);
+            for (size_t i = 0; i < noutputs; ++i)
+            {
+                rout.push_back(requested_outputs[i]);
+            }
         }
         *outputs         = new NP_NeuropodValueMap();
         (*outputs)->data = std::move(*model->model->infer(inputs->data, rout));
-        if (status != nullptr)
-        {
-            NP_ClearStatus(status);
-        }
+        NP_ClearStatus(status);
     }
     catch (std::exception &e)
     {
-        if (status != nullptr)
-        {
-            status->code    = NEUROPOD_ERROR;
-            status->message = e.what();
-        }
+        status->code    = NEUROPOD_ERROR;
+        status->message = e.what();
     }
 }
 
--- a/source/neuropod/bindings/c/c_api.cc
+++ b/source/neuropod/bindings/c/c_api.cc
@@ -18,32 +18,126 @@ limitations under the License.
 #include "neuropod/bindings/c/c_api.h"
 
 #include "neuropod/bindings/c/c_api_internal.h"
+#include "neuropod/bindings/c/np_status_internal.h"
 #include "neuropod/bindings/c/np_tensor_allocator_internal.h"
 #include "neuropod/bindings/c/np_valuemap_internal.h"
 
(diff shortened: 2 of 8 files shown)


source/neuropod/bindings/c/c_api.h
```diff
@@ -29,19 +29,44 @@ extern "C" {
 
 typedef struct NP_Neuropod NP_Neuropod;
 
-// Load a model given a path
+// Load a model given a path.
+// status is set if user provided a valid pointer.
```
VivekPanyam, line 33: I'm not sure it's a good idea for status to be optional here (and in the other methods). Doing so might make this interface somewhat error prone.

It's obviously easier to pass in `nullptr` than to use and check am `NP_Status`, but I think we should force usage.

When I was working with the TF C API, it was somewhat inconvenient to have to deal with statuses everywhere, but it forced me to do error checking and I think that led to more robust user code (and a better debugging experience).

Any thoughts @selitvin

source/neuropod/bindings/c/c_api.h
```diff
@@ -29,19 +29,44 @@ extern "C" {
...
-    // Whether or not to use out-of-process execution
-    // (using shared memory to communicate between the processes)
+    // Whether or not to use out-of-process execution.
     bool use_ope;
 
+    // These options are only used if use_ope is set to true.
+    struct NP_OPEOptions
+    {
+        bool free_memory_every_cycle;
+        char control_queue_name[256];
```
VivekPanyam, line 46: Maybe pull the 256 out into a constant?

source/neuropod/tests/test_c_api.cc
```diff
@@ -55,19 +87,30 @@ TEST(test_c_api, basic)
...
     NP_Infer(model, inputs, &outputs, status);
     if (NP_GetCode(status) != NEUROPOD_OK)
     {
-        // Throw an error here
         FAIL() << "Error from C API during inference: " << NP_GetMessage(status);
     }
 
+    // The same inference but specify requested output.
+    const char *requested_output[] = {"out"};
+    NP_InferWithRequestedOutputs(model, inputs, 1, static_cast<const char **>(requested_output), &outputs, status);
```
VivekPanyam, line 99: This static cast is unnecessary

(See https://repl.it/repls/OrangeCurlySort)

source/neuropod/bindings/c/c_api.cc
```diff
@@ -18,32 +18,126 @@ limitations under the License.
...
+                                  NP_NeuropodValueMap **     outputs,
+                                  NP_Status *                status)
+{
+    try
+    {
+        // By default empty collection of requested_ouputs expected.
+        std::vector<std::string> rout;
+        for (size_t i = 0; i < noutputs; ++i)
+        {
+            rout.push_back(requested_outputs[i]);
```
VivekPanyam, line 102: Also infer found an issue (see the failing lint check):

```

neuropod/bindings/c/c_api.cc:86: error: NULL_DEREFERENCE
--
  | pointer `null` is dereferenced by call to `NP_InferWithRequestedOutputs()` at line 86, column 5.
  | 84.   void NP_Infer(NP_Neuropod *model, const NP_NeuropodValueMap *inputs, NP_NeuropodValueMap **outputs, NP_Status *status)
  | 85.   {
  | 86. >     NP_InferWithRequestedOutputs(model, inputs, 0, nullptr, outputs, status);
  | 87.   }
  | 88.
  |  
  |  
  | Summary of the reports
  |  
  | NULL_DEREFERENCE: 1

```

I know in the example it gave, `noutputs` is 0, but we should probably add a null check on `requested_outputs` anyway

source/neuropod/bindings/c/c_api.h
vkuzmin-uber: I thought abut some extra efficiency for Infer. "error handler" instead of status may be more efficient (it would never call it on success), but let's make it required for now.

source/neuropod/bindings/c/README.md
```diff
@@ -0,0 +1,7 @@
+# The C api interface:
+
+1. To build &to unit test :
+``` 
+bazel build // neuropod/bindings/c:c_api
```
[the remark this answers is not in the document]
vkuzmin-uber: Not sure how it happened. Maybe clang-format touched it or something else happened. Anyway, fixed.

source/neuropod/bindings/c/c_api.cc
vkuzmin-uber: added checks

source/neuropod/bindings/c/c_api.h
vkuzmin-uber: Added define QUEUENAME_MAX (similar to standard FILENAME_MAX)

thuningxu:
Let's make sure we test with a C compiler.

vkuzmin-uber:
> Let's make sure we test with a C compiler.

This is a very good point. Tensorflow also uses C11 to implement smoke test for its C API. Just published update for PR, see test_c_api.c.

source/neuropod/bindings/c/c_api.h
```diff
@@ -23,25 +23,52 @@ limitations under the License.
...
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 typedef struct NP_Neuropod NP_Neuropod;
 
-// Load a model given a path
+// Load a model given a path.
+// status is required. Note that function sets it without the check on nullptr.
```
VivekPanyam, line 35: Not sure if we need this comment here and below.

In the context of our back and forth, I think it makes sense (because we made a change related to that functionality), but I think it's a little odd to include it in the final code especially since we don't document this for every param of the C API functions

source/neuropod/tests/test_c_api.c
```diff
@@ -0,0 +1,162 @@
...
+// the call to STRINGIZE(condition) in the definition of the CHECK
+// macro results in the string "foo" rather than the string "condition".
+#define STRINGIZE(expression) STRINGIZE2(expression)
+#define STRINGIZE2(expression) #expression
+
+// Like assert(), but not dependent on NDEBUG.
+#define CHECK(condition) ((condition) ? (void) 0 : CheckFailed(STRINGIZE(condition), __FILE__, __LINE__))
+#define ASSERT_EQ(expected, actual) CHECK((expected) == (actual))
+#define ASSERT_NE(expected, actual) CHECK((expected) != (actual))
+#define ASSERT_STREQ(expected, actual) ASSERT_EQ(0, strcmp((expected), (actual)))
```
VivekPanyam, line 49: If you didn't write this, can you include a comment with a link to where you got it from?

source/neuropod/bindings/c/README.md
```diff
@@ -0,0 +1,7 @@
+# The C api interface:
```
VivekPanyam, line 1: Maybe something like

```
# C API

***NOTE: This API is currently experimental***
```

source/neuropod/bindings/c/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+1. To build & to unit test:
+```
+bazel build //neuropod/bindings/c:c_api
+bazel test --test_output=errors //neuropod/tests:test_c_api
```
VivekPanyam, line 6: FYI, you don't need to run `bazel build` before running `bazel test` (it'll rebuild things as necessary)

source/neuropod/bindings/c/c_api.h
vkuzmin-uber: I will remove comment with status for now. We may think about some standard for description of public API, like Docygen or other and strictly follow it.

vkuzmin-uber: Discussed offline https://google.github.io/styleguide/cppguide.html#Function_Comments

source/neuropod/tests/test_c_api.c
vkuzmin-uber: This is pretty common implementation, but namely this one was copied from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/c/c_test.c . I add link to it.

source/neuropod/bindings/c/README.md
vkuzmin-uber: Yes, this is how I run tests. "build" is when need to produce so. btw, I am not sure if "bazel test" produces .so files really.

PR #410, opened 2020-08-07
[Experimental] Add a RuntimeOption to set inter and intra op threadpool sizes

VivekPanyam:
### Summary:

Allow users to configure intra-op and inter-op thread pool sizes for the underlying frameworks.

**Note: This API is experimental and may change in the future.**

### DIFF
--- a/source/neuropod/backends/tensorflow/tf_backend.cc
+++ b/source/neuropod/backends/tensorflow/tf_backend.cc
@@ -93,7 +93,7 @@ void check_tf_status(const tensorflow::Status &status)
 }
 
 // Get TF session options given Neuropod RuntimeOptions
-tensorflow::SessionOptions get_tf_opts(const RuntimeOptions & /*unused*/)
+tensorflow::SessionOptions get_tf_opts(const RuntimeOptions &runtime_opts)
 {
     tensorflow::SessionOptions opts;
 
@@ -102,6 +102,26 @@ tensorflow::SessionOptions get_tf_opts(const RuntimeOptions & /*unused*/)
     gpu_opts->set_allow_growth(true);
     opts.config.set_allow_soft_placement(true);
     opts.config.set_log_device_placement(false);
+
+    // Set intra and inter op parallelism
+    // See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto
+    if (runtime_opts.experimental_intra_op_parallelism_threads != 0)
+    {
+        opts.config.set_intra_op_parallelism_threads(
+            static_cast<int32_t>(runtime_opts.experimental_intra_op_parallelism_threads));
+    }
+
+    if (runtime_opts.experimental_inter_op_parallelism_threads == 1)
+    {
+        // Only use the caller thread
+        opts.config.set_inter_op_parallelism_threads(-1);
+    }
+    else if (runtime_opts.experimental_inter_op_parallelism_threads > 1)
+    {
+        // The number in runtime_opts includes the caller thread
+        opts.config.set_inter_op_parallelism_threads(
+            static_cast<int32_t>(runtime_opts.experimental_inter_op_parallelism_threads) - 1);
+    }
 
     // Note: we can't use GPUOptions::visible_device_list as it is a per process setting
     //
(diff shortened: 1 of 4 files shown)


source/neuropod/backends/tensorflow/tf_backend.cc
```diff
@@ -103,6 +103,24 @@ tensorflow::SessionOptions get_tf_opts(const RuntimeOptions & /*unused*/)
...
+    }
+
+    if (runtime_opts.inter_op_parallelism_threads == 1)
+    {
+        // Only use the caller thread
+        opts.config.set_inter_op_parallelism_threads(-1);
+    }
+    else if (runtime_opts.inter_op_parallelism_threads > 1)
+    {
+        // The numer in runtime_opts includes the caller thread
```
VivekPanyam, line 120: Typo: `number`

source/neuropod/backends/tensorflow/tf_backend.cc
```diff
@@ -93,7 +93,7 @@ void check_tf_status(const tensorflow::Status &status)
 }
 
 // Get TF session options given Neuropod RuntimeOptions
-tensorflow::SessionOptions get_tf_opts(const RuntimeOptions & /*unused*/)
+tensorflow::SessionOptions get_tf_opts(const RuntimeOptions &runtime_opts)
```
vkuzmin-uber, line 96: Not related. Some thoughts. RuntimeOption is used now in C++, C and Java API (and could be Python too). We need to keep it in sync. I have seen that Tensorflow using Proto declaration in such cases and then generates struct for languages.

We may use similar approach.

VivekPanyam: Makes sense - definitely something to look into.

I don't like TF's approach with options for their C API though. They basically require a buffer with a serialized proto as input. This makes it fairly complicated to set options directly from C.

source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -225,6 +225,21 @@ std::mutex                      loaded_op_mutex;
 TorchNeuropodBackend::TorchNeuropodBackend(const std::string &neuropod_path, const RuntimeOptions &options)
     : NeuropodBackendWithDefaultAllocator<TorchNeuropodTensor>(neuropod_path, options)
 {
+// inter and intra op parallelism settings only supported in Torch >= 1.2.0
+#if CAFFE2_NIGHTLY_VERSION >= 20190808
+    // Set intra and inter op parallelism
+    // See https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html#runtime-api
+    if (options.experimental_inter_op_parallelism_threads != 0)
+    {
+        at::set_num_interop_threads(static_cast<int32_t>(options.experimental_inter_op_parallelism_threads));
```
vkuzmin-uber, line 234: As I see, there is a torch::set_num_interop_threads that is supposed to be "public". Minor, but I think it is issue still.

vkuzmin-uber: same about other.

source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -225,6 +225,21 @@ std::mutex                      loaded_op_mutex;
...
+    }
+
+    if (options.experimental_intra_op_parallelism_threads != 0)
+    {
+        at::set_num_threads(static_cast<int32_t>(options.experimental_intra_op_parallelism_threads));
```
vkuzmin-uber, line 239: We need to give access to get_num_* somehow. User sets runtime (or not), then starts neuropod execution and should be able to check what are the settings are used, it is important for "default" case, when system sets value and also for IPE case where models share it.

source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -225,6 +225,21 @@ std::mutex                      loaded_op_mutex;
 TorchNeuropodBackend::TorchNeuropodBackend(const std::string &neuropod_path, const RuntimeOptions &options)
     : NeuropodBackendWithDefaultAllocator<TorchNeuropodTensor>(neuropod_path, options)
 {
+// inter and intra op parallelism settings only supported in Torch >= 1.2.0
+#if CAFFE2_NIGHTLY_VERSION >= 20190808
+    // Set intra and inter op parallelism
+    // See https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html#runtime-api
+    if (options.experimental_inter_op_parallelism_threads != 0)
```
vkuzmin-uber, line 232: For Interop case, it allows to set several times only for TBB case. For our case, if there are 2nd IPE model with non-zero value, it will fail. As I see in code, for non-TBB cases, it sets atomic var and I think it can be done here as well. I'd save still possibility to change it for TBB case. I am thinking about building libtorch for TBB, we have a Torchscript model/use case where TBB's "better" concurrency can be critical.

PR #411, opened 2020-08-10
[Java API] Improve docs and add experimental notice

VivekPanyam:
### Summary:
Rearranges some documentation in the README for the Java API and adds an experimental notice

See this comment for more details: https://github.com/uber/neuropod/pull/405#pullrequestreview-463963520

### Test Plan:

CI

### DIFF
--- a/source/neuropod/bindings/java/README.md
+++ b/source/neuropod/bindings/java/README.md
@@ -1,42 +1,41 @@
-# The java api interface:
+# Java API
 
-1. To build & to unit test:
-```
-bazel build //neuropod/bindings/java:neuropod_java_jar
-bazel test //neuropod/bindings/java:*
-```
+***NOTE: This API is currently experimental***
 
-2. To load a model:
-```
+## Usage
+
+```java
+// Load a model
 Neuropod model = new Neuropod(modelPath);
-```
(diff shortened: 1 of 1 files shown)


source/neuropod/bindings/java/README.md
```diff
@@ -1,42 +1,41 @@
-# The java api interface:
+# Java API
 
-1. To build & to unit test:
-```
-bazel build //neuropod/bindings/java:neuropod_java_jar
-bazel test //neuropod/bindings/java:*
```
VivekPanyam, line 6: As of #413, you need to use `./build/test.sh` or `./build/run_cpp_tests.py` to successfully run the java tests

vkuzmin-uber: I think we can specify that build and test above can be used only during development (for smoke test). But then need to perform ./build/test.sh or ./build/run_cpp_tests.py

VivekPanyam: The test command in the README will not pass anymore as of #413 without running under either of the scripts I mentioned (this is because some of the java tests now require env variables to be set by the test runner).

I'll change some of our build tools to make it easier for us to run tests for a subset of the project

PR #415, opened 2020-08-10
[Java API] Add getInput and getOutput method for Neuropod class

weijiadeng-uber:
### Summary:
Add getInputSpec and getOutputSpec method for Neuropod class

### Test Plan:

CI


source/neuropod/bindings/java/src/main/java/com/uber/neuropod/Dimension.java
```diff
@@ -65,4 +67,30 @@ public long getValue() {
...
+    @Override
+    public String toString() {
+        if (value >= 0) {
+            return String.valueOf(value);
+        } else if (value == -1) {
+            return "None";
+        } else if (value == -2) {
+            return symbol;
+        }
+        return "Not Set!";
```
VivekPanyam, line 94: Should we throw an error here? This case means the `value < -2` which likely means something went wrong

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -115,3 +117,71 @@ JNIEXPORT jstring JNICALL Java_com_uber_neuropod_Neuropod_nativeGetPlatform(JNIE
...
+        return toJavaTensorSpecList(env, outputs);
+    }
+    catch (const std::exception &e)
+    {
+        throwJavaException(env, e.what());
+    }
+    return nullptr;
+}
+
+jobject toJavaTensorSpecList(JNIEnv *env, const std::vector<neuropod::TensorSpec> &specs)
```
VivekPanyam, line 151: Can we put this in an anonymous namespace?

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -115,3 +117,71 @@ JNIEXPORT jstring JNICALL Java_com_uber_neuropod_Neuropod_nativeGetPlatform(JNIE
...
+                env->CallBooleanMethod(dims, java_util_ArrayList_add, javaDim);
+                env->DeleteLocalRef(javaDim);
+            }
+        }
+        jobject javaTensorSpec =
+            env->NewObject(com_uber_neuropod_TensorSpec, com_uber_neuropod_TensorSpec_, name, type, dims);
+        env->CallBooleanMethod(ret, java_util_ArrayList_add, javaTensorSpec);
+        env->DeleteLocalRef(name);
+        env->DeleteLocalRef(dims);
+        env->DeleteLocalRef(type);
```
VivekPanyam, line 184: We don't need to explicitly call `DeleteLocalRef` (here and in all the JNI code).

See [here](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/design.html#wp1242) in the JNI specification.

Specificially,
> Local references are valid for the duration of a native method call, and are automatically freed after the native method returns.

> In most cases, the programmer should rely on the VM to free all local references after the native method returns. However, there are times when the programmer should explicitly free a local reference. 

There are some cases where we'd want to explicitly call `DeleteLocalRef` (e.g. large loop that creates objects, references to large objects, etc), but in most cases we should be fine without. See the spec linked above for more details.

I forgot to mention this on your previous PR. If you get a chance, can you put up another PR removing all the unnecessary `DeleteLocalRef`s? Otherwise, let's create an issue to clean this up.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.h
```diff
@@ -22,13 +22,6 @@ limitations under the License.
 #ifdef __cplusplus
 extern "C" {
 #endif
-/*
- * Class:     com_uber_neuropod_Neuropod
- * Method:    nativeNew
- * Signature: (Ljava/lang/String;)J
- */
-JNIEXPORT jlong JNICALL Java_com_uber_neuropod_Neuropod_nativeNew__Ljava_lang_String_2(JNIEnv *, jclass, jstring);
-
```
VivekPanyam, line 31: Why is this removed here?

source/neuropod/bindings/java/src/main/native/jclass_register.cc
```diff
@@ -50,6 +64,35 @@ jint JNI_OnLoad(JavaVM *vm, void *reserved)
...
+        com_uber_neuropod_Dimension_symbol_ =
+            getMethodID(env, com_uber_neuropod_Dimension, "<init>", "(Ljava/lang/String;)V");
+
+        com_uber_neuropod_TensorType =
+            static_cast<jclass>(env->NewGlobalRef(findClass(env, "com/uber/neuropod/TensorType")));
+    }
+    catch (const std::exception &e)
+    {
+        throwJavaException(env, e.what());
+    }
```
VivekPanyam, line 95: Do we need to delete these GlobalRefs in `JNI_OnUnload` below?

source/neuropod/bindings/java/src/main/native/utils.cc
```diff
@@ -46,6 +48,35 @@ jclass findClass(JNIEnv *env, const char *name)
...
+
+jobject getFieldObject(JNIEnv *env, jclass dataTypes, std::string fieldName)
+{
+    jfieldID field = env->GetStaticFieldID(dataTypes, fieldName.c_str(), TENSOR_TYPE.c_str());
+    if (reinterpret_cast<jlong>(field) == 0)
+    {
+        throw std::runtime_error(std::string("Field not found: ") + fieldName);
+    }
+    return env->GetStaticObjectField(dataTypes, field);
+}
```
VivekPanyam, line 69: Should we rename this function to be more clear? It seems specific to `TENSOR_TYPE`

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTest.java
```diff
@@ -67,6 +72,39 @@ public void getPlatform() {
...
+                new TensorSpec("out", TensorType.FLOAT_TENSOR,
+                        Arrays.asList(new Dimension(-1), new Dimension(-1)))));
+        assertEquals(outputs, expected);
+    }
+
+    @Test
+    public void getOutputsTestSymbol() {
+        RuntimeOptions opts = new RuntimeOptions();
+        // opts can only be set to true for now, otherwise load backend library
+        // will fail
```
VivekPanyam, line 99: nit: `// TODO(weijiad): For now OPE is required to use the Java bindings`

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTest.java
```diff
@@ -67,6 +72,39 @@ public void getPlatform() {
...
+        opts.useOpe = true;
+        Neuropod torchModel = new Neuropod(TORCHSCRIPT_MODEL_PATH, opts);
```
VivekPanyam, line 101: Missing `torchModel.close()`. Use try-with-resources instead:

```java
try (Neuropod torchModel = new Neuropod(TORCHSCRIPT_MODEL_PATH, opts)) {
    // ...
}
```

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
weijiadeng-uber: We need to delete here, because JNI only supports max to 16 local references, and in a loop it is likely beyond this limitation.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.h
weijiadeng-uber: They are unused for a function that I have deleted. I forgot to remove this part in previous commit.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
VivekPanyam: Hmm I thought it was 512 local references, but from what I've read it's version specific.

That said, I missed that these deletes are in the loop (I thought they were at the end if the function). Let's leave them here, but in other spots where we're not in a loop, it'll probably simplify code to remove `DeleteLocalRef`

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -31,6 +31,11 @@ limitations under the License.
 
 using namespace neuropod::jni;
 
+namespace
+{
+jobject toJavaTensorSpecList(JNIEnv *env, const std::vector<neuropod::TensorSpec> &specs);
```
VivekPanyam, line 36: Can we remove the forward declaration and just move the implementation here?

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -31,6 +31,11 @@ limitations under the License.
...
+}
```
VivekPanyam, line 37: `} // namespace`

(Looks like the C++ linters aren't running on this code. I'll look into it and I'm also working on getting lint/static analysis working for the Java/JNI code)

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -115,3 +156,33 @@ JNIEXPORT jstring JNICALL Java_com_uber_neuropod_Neuropod_nativeGetPlatform(JNIE
...
+
+JNIEXPORT jobject JNICALL Java_com_uber_neuropod_Neuropod_nativeGetInputs(JNIEnv *env, jclass, jlong handle)
+{
+    try
+    {
+        auto model     = reinterpret_cast<neuropod::Neuropod *>(handle);
+        auto inputSpec = model->get_inputs();
+        return toJavaTensorSpecList(env, inputSpec);
+    }
+    catch (const std::exception &e)
```
vkuzmin-uber, line 168: #include <exception>

PR #420, opened 2020-08-18
[PythonBridge] Use an isolated relocatable python environment

VivekPanyam:
### Summary:
This PR modifies the Python Neuropod backends to be fully self-contained and isolated from the user's environment.

As part of this, it creates a `_neuropod_native_bootstrap` python package that is included alongside the backend `.so` files and the built-in python interpreter. This package is responsible for loading models and installing their python dependencies.

This PR also modifies build and coverage scripts to correctly collect coverage information when running in isolated python interpreters.

#### PR Size

This PR seems quite large `(+1107, -191)`, but many of the changes are straightforward. For example, `471` of those lines is vendoring a library in `_neuropod_native_bootstrap` (because it can't have dependencies on any python packages that are not built-in).

Also, most of this code in `_neuropod_native_bootstrap` is copied or moved in from the main Neuropod package. See the README in `source/neuropod/backends/python_bridge/_neuropod_native_bootstrap` for more details.

## Enabling isolation

The isolation behavior is **not enabled by default** as this would be a breaking change. Set the `NEUROPOD_ENABLE_PYTHON_ISOLATION` environment variable to enable it.

To start making models compatible with this behavior, ensure you are passing in `requirements` when packaging the models. These requirements should be available on the target platform version (e.g. `python 3.6`) and OS (e.g. `linux`).

### Test Plan:

Extensive local testing + CI now builds with isolation enabled


vkuzmin-uber (approved):
I will test it soon and report if find any issues.

PR #421, opened 2020-08-25
[Java API] Infer and updates to NeuropodTensor

vkuzmin-uber:
### Summary:
This is a next step in implementation of Java API that includes Inference method and necessary updates to other modules.
Also several JNI C++ methods changed to use std::string instead of "const char*".
  
### Test Plan:
Unit test covers Inference of TF addition model.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.coverage*
 *.profraw
 *.pyc
 source/compile_commands.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*.coverage*
 *.profraw
 *.pyc
 source/compile_commands.json
--- a/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
+++ b/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
@@ -76,9 +76,9 @@ public class NeuropodTensor extends NativeClass implements Serializable {
     /**
      * Flatten the tensor data and convert it to a long buffer.
      * <p>
-     * Can only be used when the tensor is INT64_TENSOR. Will trigger
-     * a copy if the tensor is created by infer method. Otherwise will
-     * not trigger a copy.
+     * Can only be used when the tensor is INT64_TENSOR.
+     *
+     * Note: It returns a buffer that is valid even after Tensor is closed.
      *
      * @return the IntBuffer
      */
@@ -95,9 +95,9 @@ public class NeuropodTensor extends NativeClass implements Serializable {
     /**
      * Flatten the tensor data and convert it to a float buffer.
      * <p>
-     * Can only be used when the tensor is FLOAT_TENSOR. Will trigger
-     * a copy if the tensor is created by infer method. Otherwise will
-     * not trigger a copy.
+     * Can only be used when the tensor is FLOAT_TENSOR.
+     *
+     * Note: It returns a buffer that is valid even after Tensor is closed.
      *
      * @return the FloatBuffer
      */
@@ -124,8 +124,9 @@ public class NeuropodTensor extends NativeClass implements Serializable {
      /**
       * Flatten the tensor data and convert it to a double buffer.
       * <p>
-      * Can only be used when the tensor is DOUBLE_TENSOR. Will trigger
-      * a copy.
+      * Can only be used when the tensor is DOUBLE_TENSOR.
+      *
+      * Note: It returns a buffer that is valid even after Tensor is closed.
       *
       * @return the DoubleBuffer
       */
--- a/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
+++ b/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
@@ -29,11 +29,21 @@ public class NeuropodTensor extends NativeClass implements Serializable {
 
     protected ByteBuffer buffer;
 
+    // This flag is used to separate NeuropodTensor that is created as Input (from java)
+    // or Output (from JNI as result of inference). This is necessary because tensor buffer
+    // is allocated differently and it is not safe to expose data directly to user.
     private final boolean isFromJava;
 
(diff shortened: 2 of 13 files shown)


vkuzmin-uber:
@VivekPanyam can we proceed with 1 failing check about coverage? I already added test to cover critical code. As I see some uncovered code is checks for sort of "impossible" cases when it tests if handle is 0 or some pointer is null (I just prefer to check and throw exception just in case). And I will improve coverage with next PRs when we get all functionality in place by adding different models (now this is only float-input TF addition).

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTest.java
```diff
@@ -122,6 +132,135 @@ public void loadModel() {
...
+
+        // Inference with requested outputs.
+        List<String> requestedOutputs = new ArrayList<String>();
+        requestedOutputs.add("out");
+        Map<String, NeuropodTensor> res2 = model.infer(inputs, requestedOutputs);
+        assertEquals(1, res2.size());
+
+        try
+        {
+           // Try failuer if provide invalid output.
```
VivekPanyam, line 249: nit: typo

VivekPanyam:
> @VivekPanyam can we proceed with 1 failing check about coverage? I already added test to cover critical code.
> 
> As I see some uncovered code is checks for sort of "impossible" cases when it tests if handle is 0 or some pointer is null (I just prefer to check and throw exception just in case) and part of JNI_OnUnload that I don't know how to cover yet.
> 
> Sure I will improve coverage with next PRs when we get all functionality in place, by adding different models (now this is only float-input TF addition). For now I want to deliver core part first and then we can work on its better coverage.

Yeah, that's fine. We also need to collect coverage information for Java code. Right now coverage only tracks C++ (including JNI) and Python so the numbers aren't entirely accurate.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
```diff
@@ -215,3 +216,65 @@ JNIEXPORT jlong JNICALL Java_com_uber_neuropod_Neuropod_nativeGetGenericAllocato
...
+            env->DeleteLocalRef(value);
+        }
+
+        auto model       = reinterpret_cast<neuropod::Neuropod *>(modelHandle);
+        auto inferredMap = model->infer(nativeMap, requestedOutputs);
+
+        // Put data to Java Map
+        auto  ret     = env->NewObject(java_util_HashMap, java_util_HashMap_);
+        auto &entries = *inferredMap;
+        for (auto &entry : entries)
```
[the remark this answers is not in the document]
vkuzmin-uber: You know, initially this was how you suggests. But I added extra reference to make sure that it doesn't create a copy (like auto always not reference except few cases).  I understand this is probably what user expect but I had no time to find what standard or some clarification says about this case and decided to add it explicitly.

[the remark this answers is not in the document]
vkuzmin-uber: I think this explains it that my fear is not right https://en.cppreference.com/w/cpp/language/range-for
```
{
auto && __range = range_expression ;
for (auto __begin = begin_expr, __end = end_expr; __begin != __end; ++__begin) 
{
range_declaration = *__begin;loop_statement
}
}
```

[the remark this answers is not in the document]
vkuzmin-uber: but at the same time it mentions some cases 
```
When used with a (non-const) object that has copy-on-write semantics, the range-based for loop may trigger a deep copy by (implicitly) calling the non-const begin() member function
that is undesirable (for instance because the loop is not actually modifying the object), it can be avoided by using std::as_const
```
not this case though

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTest.java
vkuzmin-uber: "Add unexpected "requested output" that should cause a failure."

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -77,7 +87,28 @@ public LongBuffer toLongBuffer() {
...
+        ret.rewind();
+        return ret;
+    }
+
+    /**
+     * Flatten the tensor data and convert it to a float buffer.
+     * <p>
+     * Can only be used when the tensor is FLOAT_TENSOR. Will trigger
+     * a copy if the tensor is created by infer method. Otherwise will
+     * not trigger a copy.
```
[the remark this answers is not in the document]
vkuzmin-uber: Changed to:
```
        /**
         * Flatten the tensor data and convert it to a float buffer.
         * <p>
         * Can only be used when the tensor is FLOAT_TENSOR.
         *
         * Note: It returns a buffer that is valid even after Tensor is closed.
         *
         * @return the FloatBuffer
         */
```

vkuzmin-uber:
Addressed issues. I see check lint/black is not successful that is unexpected. Is it flaky?

.gitignore
```diff
@@ -1,3 +1,4 @@
+*.coverage*
```
VivekPanyam, line 1: This ignores `.coveragerc` files and we want to keep those.

I think this might be causing your lint failure

VivekPanyam:
> Addressed issues. I see check lint/black is not successful that is unexpected. Is it flaky?

It looks like there was a new [major](https://github.com/psf/black/releases/tag/20.8b0) [release](https://github.com/psf/black/releases/tag/20.8b1) of black this morning that changed some formatting behavior. Will put up a PR to pin to the previous release

.gitignore
VivekPanyam: Can you change this @vkuzmin-uber?

vkuzmin-uber: I will revert this line completely. We can update .gitignore in separate change if necessary.

PR #422, opened 2020-08-26
[C API] NP_GetDims implementation

vkuzmin-uber:
### Summary:
Continue with basic implementation of a C API for Neuropod. Issue #407

### Test Plan:
Added test that test returned values of number of dims and data in dims array.

### DIFF
--- a/source/neuropod/bindings/c/np_tensor.cc
+++ b/source/neuropod/bindings/c/np_tensor.cc
@@ -22,9 +22,9 @@ void NP_GetDims(const NP_NeuropodTensor *tensor, size_t *num_dims, const int64_t
 {
     // get_dims returnis reference to internal tensor's dims collection.
     // Use pointer to its data that is valid as long as tensor is valid.
-    auto &dims_collection = tensor->tensor->as_tensor()->get_dims();
-    *num_dims             = dims_collection.size();
-    *dims                 = dims_collection.data();
+    const auto &dims_collection = tensor->tensor->as_tensor()->get_dims();
+    *num_dims                   = dims_collection.size();
+    *dims                       = dims_collection.data();
 }
 
 NP_TensorType NP_GetType(const NP_NeuropodTensor *tensor)
--- a/source/neuropod/bindings/c/np_tensor.cc
+++ b/source/neuropod/bindings/c/np_tensor.cc
@@ -20,9 +20,11 @@ limitations under the License.
 
 void NP_GetDims(const NP_NeuropodTensor *tensor, size_t *num_dims, const int64_t **dims)
 {
-    auto dimsCollection = tensor->tensor->as_tensor()->get_dims();
-    *num_dims           = dimsCollection.size();
-    *dims               = dimsCollection.data();
+    // get_dims return refernce to internal tensor's vector.
+    // We return a pointer to it that is valid while tensor is valid.
+    auto &dims_collection = tensor->tensor->as_tensor()->get_dims();
+    *num_dims             = dims_collection.size();
+    *dims                 = dims_collection.data();
 }
 
 NP_TensorType NP_GetType(const NP_NeuropodTensor *tensor)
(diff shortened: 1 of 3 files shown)


source/neuropod/bindings/c/np_tensor.cc
```diff
@@ -18,6 +18,13 @@ limitations under the License.
 #include "neuropod/bindings/c/np_tensor_internal.h"
 #include "neuropod/internal/neuropod_tensor_raw_data_access.hh"
 
+void NP_GetDims(const NP_NeuropodTensor *tensor, size_t *num_dims, const int64_t **dims)
+{
+    auto dimsCollection = tensor->tensor->as_tensor()->get_dims();
```
VivekPanyam, line 23: Use `auto &` here.

The `dims` pointer you're setting below is not pointing to a valid location on the heap (since `dimsCollection` will be destroyed when this function returns and the pointer returned by `data()` will no longer be valid).

The test below happens to pass because the data isn't being overwritten before you check against the expected value.

(Also, nit: snake case instead of camel case for variable names - `tensor_dims` or something instead of `dimsCollection`)

vkuzmin-uber:
I know linter shown problem and I ran autofix and didn't check it after athat! It means autofix doesn't update C file! Will look on it and update.

source/neuropod/bindings/c/np_tensor.cc
vkuzmin-uber: Yes, I always forget that auto never resolves to reference (with few exceptions)!

Do you mean, because I missed auto& it returns pointer to local - agree. With auto& it will point to what get_dims returns that return reference to tensor's internal buffer and so it should be fine, right?

VivekPanyam:
> I know linter shown problem and I ran autofix and didn't check it after athat! It means autofix doesn't update C file! Will look on it and update.

The lint script currently fails on the first linter that fails (for example if you have an error in a BUILD file, it won't even look at Python and C++). I think autofix has this behavior as well. On my todo list to fix this :) 

For now, I usually just  run `tools/lint.sh` again just to make sure everything is fine.

source/neuropod/bindings/c/np_tensor.cc
VivekPanyam: Correct, the data pointed to by `dims` is valid as long as the tensor is valid

vkuzmin-uber: Fixed.

VivekPanyam: > Do you mean, because I missed auto& it returns pointer to local - agree.

One clarification - `dims` would point to memory on the heap either way (because the data in a `vector` is on the heap).

The difference is by using `auto &` the returned data is actively managed by a vector that is still "alive" (the `dims_` vector inside NeuropodTensor).

By using `auto`, you're returning a pointer to data that was allocated and managed by `dimsCollection` but is now unused from the system's perspective.

source/neuropod/bindings/c/np_tensor.cc
```diff
@@ -18,6 +18,15 @@ limitations under the License.
 #include "neuropod/bindings/c/np_tensor_internal.h"
 #include "neuropod/internal/neuropod_tensor_raw_data_access.hh"
 
+void NP_GetDims(const NP_NeuropodTensor *tensor, size_t *num_dims, const int64_t **dims)
+{
+    // get_dims returnis reference to internal tensor's dims collection.
+    // Use pointer to its data that is valid as long as tensor is valid.
+    auto &dims_collection = tensor->tensor->as_tensor()->get_dims();
```
VivekPanyam, line 25: nit: explicity mark things as `const` if you don't intend to modify them (even though in this case `auto &` will deduce it)

tools/autofix.sh
```diff
@@ -14,7 +14,7 @@ fi
 pushd source
 
 # Reformat with clang-format
-find . -name "*.hh" -o -name "*.h" -o -name "*.cc" | xargs -L1 clang-format -style=file -i
+find . -name "*.hh" -o -name "*.h" -o -name "*.cc" -o -name "*.c" | xargs -L1 clang-format -style=file -i
```
VivekPanyam, line 17: Ah I see - I misunderstood your comment above about autofix

source/neuropod/bindings/c/np_tensor.cc
vkuzmin-uber: Yes, get_dims may have const and non-const version. Will add "const auto&"

PR #423, opened 2020-08-26
[Build] Modify run test script to filter by lang

vkuzmin-uber:
### Summary:
Now we have CC and Java tests (cc_test and java_test targets). 
Script build/run_bazel_tests.sh allows to specify --lang flag. For example:
python ../build/run_bazel_tests.py --lang java
python ../build/run_bazel_tests.py --lang cc

By default it runs tests for all langs:
python ../build/run_bazel_tests.py

### Test Plan:
Run 
build/test.sh 
python ../build/run_bazel_tests.py --lang java
python ../build/run_bazel_tests.py --lang cc


vkuzmin-uber:
> Looks good! Can we rename `run_lang_tests.py` to `run_native_tests.py` or `run_bazel_tests.py`?

run_bazel_tests looks very good.

> Also, I assume you tested with various flags locally in addition to just running `build/test.sh`, right?

Yes, I ran all combinations, including --run-gpu-tests flag too.

PR #425, opened 2020-08-31
[Java API] Improve Java Addition Model Test

vkuzmin-uber:
### Summary:
NeuropodAdditionTest can be improved to be generic test that works with TF and Torchscript addition models. Class NeuropodAdditionTest implements tests, derived classes for TF and Torchscript only prepare test environment and specify appropriate model config.

### Test Plan:
Build and then run Java tests:
python ../build/run_bazel_tests.py --lang java
and then all tests
build/test.sh

### DIFF
--- a/source/neuropod/bindings/java/src/test/java/com/uber/neuropod/DifferentFeaturesTest.java
+++ b/source/neuropod/bindings/java/src/test/java/com/uber/neuropod/DifferentFeaturesTest.java
@@ -1,0 +1,47 @@
+/* Copyright (c) 2020 UATC, LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


source/neuropod/bindings/java/src/test/java/com/uber/neuropod/DifferentFeaturesTest.java
```diff
@@ -0,0 +1,47 @@
...
+package com.uber.neuropod;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.*;
+
+import static org.junit.Assert.*;
+
+public class DifferentFeaturesTest {
```
VivekPanyam, line 25: What does this mean? Can we make the name more clear?

vkuzmin-uber: I wanted to test Named Dimensions. I decided to not create a separate class/test for that and decided to call it  Different Features Test where I'd put some tests that can't be organized into some domain. I guess once this class gets more tests that can be combined together, I will extract them into separate Test.

Consider that intention, any idea how I could call it or use other approach?

vkuzmin-uber: I am thinking about renaming it to TesnorSpecTest that reflects a current test and I will be able to use it for all related to spec.

VivekPanyam (approved):
LGTM!

PR #426, opened 2020-09-01
[Java API] Complete NeuropodTensor and Allocator implementation

vkuzmin-uber:
### Summary:
This PR includes complete implementation of Neuropod Tensor and Allocator classes, including support of STRING_TENSOR.

### Test Plan:
CI + Added additional tests

### DIFF
--- a/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
+++ b/source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
@@ -17,6 +17,7 @@ package com.uber.neuropod;
 
 import java.io.Serializable;
 import java.nio.*;
+import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -87,7 +88,7 @@ public class NeuropodTensor extends NativeClass implements Serializable {
         if (isFromJava) {
             return buffer.asLongBuffer();
         }
-        LongBuffer ret = LongBuffer.allocate((int)getNumberOfElements()).put(buffer.asLongBuffer());
+        LongBuffer ret = LongBuffer.allocate((int) getNumberOfElements()).put(buffer.asLongBuffer());
         ret.rewind();
         return ret;
     }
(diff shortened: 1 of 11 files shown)


vkuzmin-uber:
@VivekPanyam the code is not complete yet and this is why I limited with code coverage here. Next PR will bring rest of Allocator and more tests for String Tensor.

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTest.java
```diff
@@ -201,6 +196,22 @@ public void InputLongTensor() {
        allocator.close();
     }
 
+    @Test
+    public void InputStringTensor() {
+       NeuropodTensorAllocator allocator = model.getTensorAllocator();
+
+       TensorType type = TensorType.STRING_TENSOR;
+
+       // Note that for String Tensor we can't use low level allocator's method tensorFromMemory
```
vkuzmin-uber, line 205: You may see, we can't allocate String Tensor yet and this is why I can't cover other methods from this review. I don't want to use some hacks and instead next PR will bring necessary functionality and test.

vkuzmin-uber:
Need to rebase it after other PRs merged.

vkuzmin-uber:
After rebase it has now Java code coverage!

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_NeuropodTensor.cc
```diff
@@ -109,3 +112,90 @@ JNIEXPORT jlong JNICALL Java_com_uber_neuropod_NeuropodTensor_nativeGetNumberOfE
...
+        default:
+            // Here copy data twice
+            const auto &elementList = typedTensor->get_data_as_vector();
+            for (const auto &elem : elementList)
+            {
+                jstring convertedElem = env->NewStringUTF(elem.c_str());
+                env->CallBooleanMethod(ret, java_util_ArrayList_add, convertedElem);
+                env->DeleteLocalRef(convertedElem);
+            }
+        }
```
VivekPanyam, line 172: Can we improve this? Maybe we should just add this as a friend function of `TypedNeuropodTensor<std::string>`. That way, it could have individual element access without a lot of the messiness here.

Also, maybe it's worth benchmarking just always using `get_data_as_vector`.  Each elementwise access to a string tensor goes through virtual calls which may be slow. For larger tensors, it's possible/likely that the branch predictor will do a good job and minimize the impact, but still worth benchmarking.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_NeuropodTensor.cc
```diff
@@ -109,3 +112,90 @@ JNIEXPORT jlong JNICALL Java_com_uber_neuropod_NeuropodTensor_nativeGetNumberOfE
...
+                                                                                jlong index,
+                                                                                jlong handle)
+{
+    try
+    {
+        auto stringTensor = (*reinterpret_cast<std::shared_ptr<neuropod::NeuropodValue> *>(handle))
+                                ->as_tensor()
+                                ->as_typed_tensor<std::string>();
+        const auto &       strList = stringTensor->get_data_as_vector();
+        const std::string &elem    = strList[index];
```
VivekPanyam, line 193: This is super inefficient if we just need one element (because it makes a copy of the entire tensor). Similar to above, let's just add this as a friend function of `TypedNeuropodTensor<std::string>` so we can just get the single element

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -106,110 +107,187 @@ public FloatBuffer toFloatBuffer() {
...
     private void checkType(TensorType type) {
         if (getTensorType() != type) {
-            throw new NeuropodJNIException("TensorType mismatch! Expected " +
+            throw new NeuropodJNIException("tensorType mismatch, expected " +
                     type.name() + ", found " + getTensorType().name());
 
         }
     }
 
+    private long toPos(long[] index) {
```
VivekPanyam, line 268: Please add test coverage for this (using tensors with somewhat complex dimensions)

VivekPanyam (changes requested):
It looks like a large portion of this PR is not covered by tests.

Especially since parts of the change could be error-prone (e.g. multidimensional indices to flat index), I think we should include some additional tests before landing. Some other basic tests that at least execute the happy path of the changes in this PR would be good as well.

Also, I think we could greatly simplify the string tensor handling (and get rid of some utility functions) if we just add some friend functions of `TypedNeuropodTensor<std::string>`.

Instead of making these friend functions specific to the Java API, maybe we should just extend https://github.com/uber/neuropod/blob/master/source/neuropod/internal/neuropod_tensor_raw_data_access.hh to add a friend `NeuropodStringTensorRawDataAccess` struct or something.

VivekPanyam:
Yeah, it looks like only ~5.42% of this PR is covered by tests. We need to improve that significantly before we can land this PR.

vkuzmin-uber:
> It looks like a large portion of this PR is not covered by tests.
> 
> Especially since parts of the change could be error-prone (e.g. multidimensional indices to flat index), I think we should include some additional tests before landing. Some other basic tests that at least execute the happy path of the changes in this PR would be good as well.
> 
> Also, I think we could greatly simplify the string tensor handling (and get rid of some utility functions) if we just add some friend functions of `TypedNeuropodTensor<std::string>`.
> 
> Instead of making these friend functions specific to the Java API, maybe we should just extend https://github.com/uber/neuropod/blob/master/source/neuropod/internal/neuropod_tensor_raw_data_access.hh to add a friend `NeuropodStringTensorRawDataAccess` struct or something.

vkuzmin-uber:
I closed it by mistake. Reopen.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
vkuzmin-uber: Yes, I will add necessary code to this PR and add test coverage that covers toPos.

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_NeuropodTensor.cc
vkuzmin-uber: I considered a little but different approach. I added to TypedNeuropodTensor<std::string>:

get_data_at(index)
set_data_at(index, value)

that allows to avoid dealing with intermediate std::vector.

I also removed mapStringTensor+accessor because having get_set_at looks like optimal enough. 

Please look and advise

source/neuropod/bindings/java/src/main/native/com_uber_neuropod_NeuropodTensor.cc
vkuzmin-uber: Added get_data_at(index). Let me know what you think about this approach.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
vkuzmin-uber: Before I was going to create a last PR with Tensor Allocator that had all functionality and so Unit tests. I decided to add it and Unit Tests to this PR.

PR will become slightly bigger (though I simplified code with get_data_at/set_data_at and removed some code).  But this way it will be more consistent and we will see what is covered and what missed better in this PR.

vkuzmin-uber:
> It looks like a large portion of this PR is not covered by tests.

I added rest if Tensor Allocator implementation and tests that I was going to publish as a last PR. Though I removed some code as you advise (with getter/helper at String Tensor) and hope now PR will look more consistent.

I will wait for build and coverage result to see what coverage missed and address in next update if necessary.

source/neuropod/bindings/java/src/main/java/com/uber/neuropod/NeuropodTensor.java
```diff
@@ -87,7 +88,7 @@ public LongBuffer toLongBuffer() {
         if (isFromJava) {
             return buffer.asLongBuffer();
         }
-        LongBuffer ret = LongBuffer.allocate((int)getNumberOfElements()).put(buffer.asLongBuffer());
+        LongBuffer ret = LongBuffer.allocate((int) getNumberOfElements()).put(buffer.asLongBuffer());
```
vkuzmin-uber, line 91: @VivekPanyam 

You can see some parts of to<Type>Buffer are not covered (only toFloatBuffer). 

This is because there is an Addition Test with model that has only Float Input and Output tensors.

Current code can be executed only if we have Output Tensor and call to<Type>Buffer(). We can create Output Buffer only using "infer" call. I don't want to add some hack that would allow us to create Output Tensors of different types only for test.

Instead, I will add test with Model that covers all different types (we use Addition model with all possible types to test our Go service with neuropod and I will use it as a base).

If you are ok with that, we can proceed with this PR and I will continue to work on test coverage.

vkuzmin-uber (commented):
@VivekPanyam  I left a comment in code about uncovered code in to<Type>Buffer methods. Please look and let me know what you think.

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodAdditionTest.java
```diff
@@ -121,75 +120,6 @@ public void loadModel() {
         }
     }
 
-    @Test
-    public void InputDoubleTensor() {
```
sepo-eng, line 125: Why are we removing these tests?

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodTensorAllocatorTest.java
```diff
@@ -16,44 +16,307 @@
...
-        TensorType type = TensorType.INT64_TENSOR;
-        ByteBuffer buffer = ByteBuffer.allocateDirect(4 * type.getBytesPerElement()).order(ByteOrder.nativeOrder());
-        LongBuffer longBuffer = buffer.asLongBuffer();
+    public void getTensorValueAsIntWithOutOfBounds() {
+        IntBuffer intBuffer = IntBuffer.allocate(4);
+        for (int i = 0; i < 4; i++) {
+            intBuffer.put(i);
+        }
+        intBuffer.rewind();
+        long[] shape = new long[] {2, 2};
```
sepo-eng, line 45: It might be useful to add tests with different shapes.

sepo-eng (commented):
LGTM

source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodAdditionTest.java
vkuzmin-uber: I moved (and refactored) all tests that not related to AdditionTest to source/neuropod/bindings/java/src/test/java/com/uber/neuropod/NeuropodAdditionTest.java , you may see it transformed into tensorFromMemory* tests

VivekPanyam (commented):
Thanks for adding test coverage!

I didn't take a super in-depth look this time (especially at the test implementations), but looks good overall.

Let me know what you think about `flat`. I can put up a PR or feel free to implement it yourself if that's easier. Since you'll need to rebase anyway (as we talked about offline), it might not be too much overhead to just rebase on an PR that implements `flat`

source/neuropod/internal/neuropod_tensor.hh
```diff
@@ -502,6 +502,44 @@ public:
...
+        const auto numel = get_num_elements();
+        if (index >= numel)
+        {
+            NEUROPOD_ERROR_HH("The index is out of bounds in the tensor.");
+        }
+
+        // Set element by index.
+        set(index, value);
+    }
+
```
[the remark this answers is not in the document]
vkuzmin-uber: Is it some pattern flat instead of get/set? I am not familiar with it but anyway, you can publish it and I will be looking on it through review. 

I am ok to wait for this and see how it solves problem for StringTensor, I need to rebase anyway.

VivekPanyam (approved):
LGTM

This PR is fairly large and, as I've reviewed it a few times already, I didn't take a very in-depth look this time. I think this is okay because the Java API is experimental and it seems like @sepo-eng also took a look.

@sepo-eng do you have any other thoughts or should we go ahead and merge it?

sepo-eng (approved):
I don't have anything to add @VivekPanyam.

Our tests still only cover a single shape, and it would be good to test others as well to ensure the logic works for the general case. @vkuzmin-uber, I'll sign off, and leave it to you to determine if this is necessary, or if you want to include them in a different PR.

vkuzmin-uber:
Thanks guys. Sean, I decided to not add more to this *huge* PR. And also we may need to discuss test coverage for *bindings*. 

We cover everything possible in C++/Python API including many  all edge cases. Should we do the same for bindings Java, C, Python? I feel like having the same tests (data sets) on every other language doesn't scale. Such unit test code maintenance, test execution time/resources has low value if compare with benefits.

Interesting to see what approach TF uses

In our case, it should be combination of:

1) Must have coverage of C++ API + test of Python bindings can replace some cases where this is reasonable because we know that Python is light wrapper for C++.

2) Coverage for Java, C (Go next) - can be like a smoke tests and to reach goal coverage goal but not try to repeat 1) untill we see some solution of reusable tests.

VivekPanyam:
Okay, I'll go ahead and merge this.

I agree that we don't need to repeat coverage of the entire core library through the bindings, but we should ensure that we have code coverage for everything in the bindings themselves. There are a lot of ways to do this (as you alluded to) so we can chat more offline and come up with a good strategy for this.

This isn't blocking because the Java API is currently experimental so we don't have to be _as_ strict about this for now.

PR #428, opened 2020-09-02
Configure code coverage for Java code

vkuzmin-uber:
### Summary:
Added "bazel coverage" call that creates LCOV combined coverage report for Java code.
Neuropod has integration with codecov service that supports LCOV TXT format.

### Test Plan:
Create PR, update Java file and check if codecov can handle it.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -66,7 +66,7 @@ if [[ $@ == *'--config=coverage'* ]]; then
     # generate coverage for Java
     pushd source
     # Set PATH for Java tests
-    export PATH=$PATH:`pwd`/bazel-bin/neuropod/multiprocess/
-    bazel coverage --collect_code_coverage --instrument_test_targets --instrumentation_filter='/java[/:]' --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main //neuropod/bindings/java/...
+    PATH=$PATH:`pwd`/bazel-bin/neuropod/multiprocess/
+    bazel coverage --collect_code_coverage --instrumentation_filter='/java[/:]' --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main //neuropod/bindings/java/...
     popd
 fi
--- a/build/build.sh
+++ b/build/build.sh
@@ -60,3 +60,13 @@ if [[ $(uname -s) == 'Linux' ]]; then
 fi
 
 popd
+
+# Use a naive substring searching to check if coverage is requested.
+if [[ $@ == *'--config=coverage'* ]]; then
+    # generate coverage for Java
+    pushd source
+    # Set PATH for Java tests
+    export PATH=$PATH:`pwd`/bazel-bin/neuropod/multiprocess/
+    bazel coverage --collect_code_coverage --instrument_test_targets --instrumentation_filter='/java[/:]' --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main //neuropod/bindings/java/...
+    popd
+fi
--- a/build/build.sh
+++ b/build/build.sh
@@ -60,13 +60,3 @@ if [[ $(uname -s) == 'Linux' ]]; then
 fi
 
 popd
-
-# Use a naive substring searching to check if coverage is requested.
-if [[ $@ == *'--config=coverage'* ]]; then
-    # generate coverage for Java
-    pushd source
-    # Set PATH for Java tests
-    PATH=$PATH:`pwd`/bazel-bin/neuropod/multiprocess/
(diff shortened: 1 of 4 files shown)


source/.bazelrc
```diff
@@ -21,7 +21,7 @@ build --cxxopt='-D_GLIBCXX_USE_CXX11_ABI=0'
 build --extra_toolchains=@bazel_tools//tools/python:autodetecting_toolchain_nonstrict
 
 # Build with coverage info
-build:coverage --copt='-fprofile-instr-generate' --copt='-fcoverage-mapping' --linkopt='-fprofile-instr-generate' --linkopt='-fcoverage-mapping'
+build:coverage --collect_code_coverage --instrument_test_targets --instrumentation_filter="/java[/:]" --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main --copt='-fprofile-instr-generate' --copt='-fcoverage-mapping' --linkopt='-fprofile-instr-generate' --linkopt='-fcoverage-mapping'
```
VivekPanyam, line 24: Why do we need these flags here and in `build/build.sh`?

build/build.sh
```diff
@@ -60,3 +60,13 @@ if [[ $(uname -s) == 'Linux' ]]; then
...
+
+# Use a naive substring searching to check if coverage is requested.
+if [[ $@ == *'--config=coverage'* ]]; then
+    # generate coverage for Java
+    pushd source
+    # Set PATH for Java tests
+    PATH=$PATH:`pwd`/bazel-bin/neuropod/multiprocess/
+    bazel coverage --collect_code_coverage --instrumentation_filter='/java[/:]' --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main //neuropod/bindings/java/...
+    popd
+fi
```
VivekPanyam, line 72: Can we move this to `build/test.sh` and `build/test_gpu.sh`? We don't want the build script to be running tests

build/coverage.sh
```diff
@@ -29,3 +29,9 @@ bazel query 'kind("cc_binary|cc_test", ...)' | sed 's/\/\//-object bazel-bin\//g
 bazel query 'kind("cc_binary|cc_test", ...)' | sed 's/\/\//-object bazel-bin\//g' |  sed 's/:/\//g' | paste -sd ' ' | xargs ./bazel-source/external/llvm_toolchain/bin/llvm-cov show -instr-profile=/tmp/neuropod_coverage/code.profdata -ignore-filename-regex="(external|tests)" > coverage.txt
 sed -i 's+/proc/self/cwd/+source/+g' coverage.txt
 popd
+
+# bazel coverage configured to create a combined LCOV coverage report for java code.
+# codecov supports Lcov TXT and so .txt file. Copy and rename it acc:ordingly.
```
VivekPanyam, line 34: nit: typo

build/build.sh
VivekPanyam: Also, we want to add a `--nocache_test_results` here to make sure that bazel actually is running all the tests

VivekPanyam (commented):
Nice! This should be helpful :)

build/build.sh
```diff
@@ -60,3 +60,13 @@ if [[ $(uname -s) == 'Linux' ]]; then
...
 
 popd
+
+# Use a naive substring searching to check if coverage is requested.
+if [[ $@ == *'--config=coverage'* ]]; then
+    # generate coverage for Java
+    pushd source
+    # Set PATH for Java tests
+    PATH=$PATH:`pwd`/bazel-bin/neuropod/multiprocess/
+    bazel coverage --collect_code_coverage --instrumentation_filter='/java[/:]' --combined_report=lcov --coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main //neuropod/bindings/java/...
```
VivekPanyam, line 70: You can replace all these flags with a `--config=coverage` to use the flags you added to `.bazelrc`

source/.bazelrc
vkuzmin-uber: I thought I reverted it back. It doesn't work as I expected really.

build/build.sh
vkuzmin-uber: I need to revert .bazelrc. This is what I tried first, I thought that bazel build --config=coverage would work as "bazel coverage" but it didn't. So, my plan - revert .bazelrc and use this command line directly.

build/build.sh
vkuzmin-uber: Maybe move it to build/coverage.h ? In test, I would need to add some flag to not run "bazel coverage" when user only want to run tests.

vkuzmin-uber: I will test if nocache_test_results works in this case

vkuzmin-uber:
I tested latest change with build/docker_build.sh and see it generated report for Java files only w/o Tests as we need it. 
Lets see result in buildkite/codcov

VivekPanyam (approved):
LGTM!

PR #429, opened 2020-09-02
Fix Java API example

vkuzmin-uber:
### Summary:
Example incorrectly demonstrated using of a Java try-with block for Tensor allocations. 
Fix it and leave note that try-with is safe for multiple allocations even if one of it throws exception. 

### Test Plan:
N/A

### DIFF
--- a/source/neuropod/bindings/java/README.md
+++ b/source/neuropod/bindings/java/README.md
@@ -42,5 +42,5 @@ model.close();
 ```
 
 Note: Using [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) can help improve exception safety of your code.
-This is usefull in case of multiple Tensor allocations because:
-"Resources are initialized in left-to-right order. If a resource fails to initialize (that is, its initializer expression throws an exception), then all resources initialized so far by the try-with-resources statement are closed."
+This is useful in case of multiple Tensor allocations because:
+"Resources are initialized in left-to-right order. If a resource fails to initialize (that is, its initialize expression throws an exception), then all resources initialized so far by the try-with-resources statement are closed."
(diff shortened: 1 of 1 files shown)


source/neuropod/bindings/java/README.md
```diff
@@ -5,37 +5,42 @@
...
-// Clean up
+// Clean up other Neuropod resources.
 allocator.close();
 output.close();
-valueMap.close();
 model.close();
 ```
 
 Note: Using [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) can help improve exception safety of your code.
+This is usefull in case of multiple Tensor allocations because:
```
VivekPanyam, line 45: typo: `useful`

vkuzmin-uber: Thanks! did spellcheck in vim this time.

vkuzmin-uber:
Not sure how it happened that codecov thinks the change in README reduces coverage. Probably this is something on codecov side.

source/neuropod/bindings/java/README.md
```diff
@@ -5,37 +5,42 @@
...
+// Clean up other Neuropod resources.
 allocator.close();
 output.close();
-valueMap.close();
 model.close();
 ```
 
 Note: Using [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) can help improve exception safety of your code.
+This is useful in case of multiple Tensor allocations because:
+"Resources are initialized in left-to-right order. If a resource fails to initialize (that is, its initialize expression throws an exception), then all resources initialized so far by the try-with-resources statement are closed."
```
VivekPanyam, line 46: typo: `its initializer expression` (instead of initialize)

VivekPanyam (commented):
I think the spellcheck changed the quote.


Also would you mind linking to the place you got the quote at the end from?

source/neuropod/bindings/java/README.md
vkuzmin-uber: yes, my Vim spellchecker suggested to change. This is a quote from here https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3 . I will return it back and this link.

VivekPanyam (approved):
LGTM!

PR #433, opened 2020-09-16
Fix a bug with string tensors and OPE

### LINKED ISSUE #397: Crash with "signal SIGSEGV: segmentation violation code" with multiple OPE instances
## Bug

Master process crash with "signal SIGSEGV: segmentation violation code" when uses multiple OPE instances of simple "string python" model, under high load 

signal SIGSEGV: segmentation violation code

SIGSEGV happens inside "neuropod::tensor_from_id(std::__1::array<char, 24ul> const&)"

It doesn't happen immediately. There are particular conditions - high volume/load, particular number of concurrent callers.

See below details with Stack and Trace logs:

## To Reproduce

"Master" Process with 4 OPE instances of the same "string_python" mode performs round-robin on incoming Inference requests. 4 concurrent callers perform Inference calls. Master crashes with SEGV not immediately but at "high volume". It is 100% reproducible if generate 20K calls. If I send per 2K calls, it can take many iterations to hit it.  

Enabled TRACE logs

```
07/14/20 21:02:41.269865: T ./neuropod/multiprocess/mq/ipc_message_queue_impl.hh:152] [thread 118825, process 118433] OPE: Sending IPC control message 2.
07/14/20 21:02:41.269898: T ./neuropod/multiprocess/mq/ipc_message_queue_impl.hh:152] [thread 118442, process 118433] OPE: Sending IPC control message 2.
07/14/20 21:02:41.269912: T ./neuropod/multiprocess/mq/ipc_message_queue_impl.hh:91] [thread 118735, process 118733] OPE: Read thread received IPC control message 2.
07/14/20 21:02:41.269922: T neuropod/multiprocess/mq/transferrables.cc:56] [thread 118735, process 118733] OPE: Clearing transferrables for msg with id 1032
fatal error: unexpected signal during runtime execution
07/14/20 21:02:41.269957: T ./neuropod/multiprocess/mq/ipc_message_queue_impl.hh:91] [thread 118597, process 118595] OPE: Read thread received IPC control message 2.
07/14/20 21:02:41.269975: T neuropod/multiprocess/mq/transferrables.cc:56] [thread 118597, process 118595] OPE: Clearing transferrables for msg with id 1033
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7fd320c10480 pc=0x7fd320d43797]
``` 
With ASAN

```
==27092==Register values:
rax = 0x000060300025ab48  rbx = 0x000060700005aae0  rcx = 0x00000000000000c4  rdx = 0x00000000000000c5
rdi = 0x0000000011b87060  rsi = 0x0000000011b870e0  rbp = 0x000070000b7e29f0  rsp = 0x000070000b7e2920
 r8 = 0x0000000000000098   r9 = 0x0000000011e41440  r10 = 0x0000000000000011  r11 = 0x0000000011e41458
r12 = 0x0000000000000000  r13 = 0x0000000011e414b8  r14 = 0x000060e000130ca0  r15 = 0x0000000000000000
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0xbac7d) in neuropod::tensor_from_id(std::__1::array<char, 24ul> const&)
```

Stack
```
2020-07-07T02:53:31.656-0700	DEBUG	model/models.go:153	Tensor: &Tensor{Value:[],Dimensions:[1 1],Dtype:DTYPE_STRING,StringVector:[benchmark],}
    #0 0x5f01c7d in neuropod::tensor_from_id(std::__1::array<char, 24ul> const&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0xbac7d)
    #1 0x5f033db in void neuropod::ipc_deserialize<std::__1::shared_ptr<neuropod::NeuropodValue> >(std::__1::basic_istream<char, std::__1::char_traits<char> >&, std::__1::shared_ptr<neuropod::NeuropodValue>&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0xbc3db)
    #2 0x5ef9266 in void neuropod::ipc_deserialize<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue> >(std::__1::basic_istream<char, std::__1::char_traits<char> >&, std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue>, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::shared_ptr<neuropod::NeuropodValue> > > >&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0xb2266)
    #3 0x5ef900f in void neuropod::detail::deserialize_payload<std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue>, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::shared_ptr<neuropod::NeuropodValue> > > >, neuropod::MessageType>(neuropod::detail::WireFormat<neuropod::MessageType> const&, std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue>, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::shared_ptr<neuropod::NeuropodValue> > > >&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0xb200f)
    #4 0x5ee67cb in neuropod::(anonymous namespace)::MultiprocessNeuropodBackend::infer_internal(std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue>, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::shared_ptr<neuropod::NeuropodValue> > > > const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0x9f7cb)
    #5 0x5f3f08f in neuropod::NeuropodBackend::infer(std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue>, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::shared_ptr<neuropod::NeuropodValue> > > > const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0xf808f)
    #6 0x5e4c340 in neuropod::Neuropod::infer(std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::shared_ptr<neuropod::NeuropodValue>, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::shared_ptr<neuropod::NeuropodValue> > > > const&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&) (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/lib/libneuropod.so:x86_64+0x5340)
    #7 0x494a31a in Infer (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/./ma-dl-inference:x86_64+0x494a31a)
    #8 0x4948d07 in _cgo_f1624c697bd0_Cfunc_Infer (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/./ma-dl-inference:x86_64+0x4948d07)
    #9 0x406918f in runtime.asmcgocall (/Users/vkuzmin/gocode/src/code.uber.internal/data/michelangelo-deeplearning-inference/./ma-dl-inference:x86_64+0x406918f)
```

## Environment

 - Neuropod Version (e.g., 0.2.0): 0.2.0
 - OS (Linux, macOS): Linux and macOS
 - Language (Python, C++, Go bindings): Go binding
 - Python version: 
 - Using OPE: yes
 
If this bug report is about running a specific model:

 - Neuropod backend (e.g. TensorFlow, PyTorch, Keras, TorchScript, Python): Python
 - Framework version (e.g. 1.3.0): 2.7

If running on GPU:

 - CUDA/cuDNN version:
 - The output of `nvidia-smi`:
 ```
 
 ```
 
 - Any other relevant information:

Python Model is simple:
```
  class Model:
      def __call__(self, str_input):
          return {'str_output': str_input}

  def get_model(data_root):
      return Model()
```

## Additional context
It fails if there are 4 OPE instances and 4 concurrent callers.
It doesn't fail if I use  1, 2, 8 concurrent clients.
It fails if I send 20K messages, when it reaches ~3-4K. If I send by batches per 2 K, it may fails around 8K number.

VivekPanyam:
### Summary:

#397 describes a segfault that happens in certain cases when creating several string tensors and using OPE with multiple models across threads.

This PR implements a test that can reproduce the segfault a very high percentage of the time along with a fix for the root cause of the issue.

This line was the root cause of the issue:

https://github.com/uber/neuropod/blob/14a7e67edfe61930e123582be26075b9efcbe8f7/source/neuropod/multiprocess/shm_tensor.hh#L195-L196

We were not storing the unique_ptr created by `make_unique` so the destructor was immediately called. The reason this didn't cause visible issues in all cases has to do with how caching of shared memory blocks is implemented.

When `block` is destroyed in the code above, the underlying shared memory block is returned to an in-process cache as it's likely that the block will be loaded again. As such, the SHM block is not actually unmapped so the address returned by `get_raw_data_ptr()` still points to a valid location.

In #397, it was observed that the issue only happens when running inference in parallel across multiple threads and using the default setting of `free_memory_every_cycle = true;`. This is because such a configuration significantly increases the likelihood of clearing a shared memory cache (and unmapping the blocks in the cache) while another thread is accessing data in one of those shared memory blocks.

Fixes #397.

### Test Plan:

Added a test that consistently failed before the fix, but passes on every run after the fix.

### DIFF
--- a/source/neuropod/multiprocess/shm_tensor.hh
+++ b/source/neuropod/multiprocess/shm_tensor.hh
@@ -192,9 +192,10 @@ public:
                       const SHMBlockID &          block_id)
         : TypedNeuropodTensor<std::string>(copy_and_strip_last_dim(dims)), write_buffer_(this->get_num_elements())
     {
-        auto base_ptr =
-            stdx::make_unique<SHMNeuropodTensor<uint8_t>>(dims, std::move(block), data, block_id)->get_raw_data_ptr();
-        auto max_len = dims[dims.size() - 1];
+        auto byte_block = stdx::make_unique<SHMNeuropodTensor<uint8_t>>(dims, std::move(block), data, block_id);
+
+        auto base_ptr = byte_block->get_raw_data_ptr();
+        auto max_len  = dims[dims.size() - 1];
 
         // Copy into our local buffer
         auto numel = get_num_elements();
--- a/source/neuropod/tests/test_ope_multiple_instances.cc
+++ b/source/neuropod/tests/test_ope_multiple_instances.cc
@@ -1,0 +1,79 @@
+/* Copyright (c) 2020 UATC, LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "neuropod/neuropod.hh"
+#include "neuropod/tests/ope_overrides.hh"
+
+#include <thread>
+
+TEST(test_ope_multiple_instances, multithreaded)
+{
+    constexpr auto NUM_INSTANCES = 4;
+    constexpr auto NUM_REQUESTS  = 2048;
+
+    // Create models (on the main thread)
+    std::vector<neuropod::Neuropod> models;
+    for (size_t i = 0; i < NUM_INSTANCES; i++)
+    {
+        neuropod::RuntimeOptions opts;
+        opts.use_ope = true;
+        models.emplace_back(
+            "neuropod/tests/test_data/pytorch_strings_model/", detail::ope_backend_location_overrides, opts);
+    }
+
+    // Start worker threads to run inference
+    std::vector<std::thread> workers;
+    for (size_t idx = 0; idx < models.size(); idx++)
+    {
+        workers.emplace_back([idx, &models]() {
+            // Every thread uses a dedicated instance
+            auto &model = models.at(idx);
+
+            // Shape and target information to validate output
+            const std::vector<int64_t>     shape  = {3};
+            const std::vector<std::string> target = {"apple sauce", "banana pudding", "carrot cake"};
+
+            for (size_t i = 0; i < NUM_REQUESTS; ++i)
+            {
+                auto x_ten = model.allocate_tensor<std::string>(shape);
+                auto y_ten = model.allocate_tensor<std::string>(shape);
+
+                x_ten->copy_from({"apple", "banana", "carrot"});
+                y_ten->copy_from({"sauce", "pudding", "cake"});
+
+                // Run inference
+                const auto output_data = model.infer({{"x", x_ten}, {"y", y_ten}});
+
+                // Check that the output data matches expected
+                const auto out_tensor = output_data->at("out")->as_typed_tensor<std::string>();
+
+                const auto out_vector = out_tensor->get_data_as_vector();
+                const auto out_shape  = out_tensor->get_dims();
+
+                // Check that the output data matches
+                EXPECT_EQ(out_vector.size(), 3);
+                EXPECT_TRUE(out_vector == target);
+
+                // Check that the shape matches
+                EXPECT_TRUE(out_shape == shape);
+            }
+        });
+    }
+
+    // Join threads
+    std::for_each(workers.begin(), workers.end(), [](std::thread &t) { t.join(); });
+}
--- a/source/neuropod/tests/BUILD
+++ b/source/neuropod/tests/BUILD
@@ -365,6 +365,25 @@ cc_test(
 )
 
 cc_test(
+    name = "test_ope_multiple_instances",
+    srcs = [
+        "test_ope_multiple_instances.cc",
+    ],
+    data = [
+        "//neuropod/backends/python_bridge:libneuropod_pythonbridge_backend.so",
+    ],
+    tags = [
+        "no_trace_logging",
+        "requires_path",
+    ],
+    deps = [
+        ":neuropod_test_utils",
+        "//neuropod:neuropod_impl",
+        "//neuropod/multiprocess",
+    ],
+)
+
+cc_test(
     name = "test_internal_neuropod_tensor",
     srcs = [
         "test_internal_neuropod_tensor.cc",


source/neuropod/tests/test_ope_multiple_instances.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "gtest/gtest.h"
+#include "neuropod/neuropod.hh"
+#include "neuropod/tests/ope_overrides.hh"
+
+#include <thread>
+
+TEST(test_ope_multiple_instances, multithreaded)
+{
+    const auto NUM_INSTANCES = 4;
+    const auto NUM_REQUESTS  = 2048;
```
vkuzmin-uber, line 26: I'd declare both as size_t explicitly (size_t is a right type for num/size/index in 64bit world) and change then iterator to size_t in a for loop below:
```
for (unsigned long i = 0; i < NUM_INSTANCES; ++i)

File: source/neuropod/tests/test_ope_multiple_instances.cc [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+            for (size_t i = 0; i < NUM_REQUESTS; ++i)
+            {
+                auto x_ten = model.allocate_tensor<std::string>({3});
+                auto y_ten = model.allocate_tensor<std::string>({3});
+
+                x_ten->copy_from({"apple", "banana", "carrot"});
+                y_ten->copy_from({"sauce", "pudding", "cake"});
+
+                // Run inference
+                const auto output_data = model.infer({{"x", x_ten}, {"y", y_ten}});
```
vkuzmin-uber, line 55: It makes sense to test output_data values in this test, theoretically it may catch some other bugs multithread/concurrent mode with large number of requests.

source/neuropod/tests/test_ope_multiple_instances.cc
```diff
@@ -0,0 +1,62 @@
...
+        neuropod::RuntimeOptions opts;
+        opts.use_ope = true;
+        models.emplace_back(
+            "neuropod/tests/test_data/pytorch_strings_model/", detail::ope_backend_location_overrides, opts);
+    }
+
+    // Start worker threads to run inference
+    std::vector<std::thread> workers;
+    for (size_t idx = 0; idx < models.size(); ++idx)
+    {
```
vkuzmin-uber, line 41: I don't remember if this is necessary in C++ but definitely a problem in Go's goroutine and other languages: 

Copy idx to local var before sending it to closure otherwise it may happen that all threads runs with latest idx. I'd recommend to use this trick to avoid confusion (or comment please if you 100% sure that this is not a problem here)

 auto modelIndex = idx;

vkuzmin-uber (commented):
Looks good. Left minor comments. Ready to approve after you look and decide if it is worth change.

source/neuropod/tests/test_ope_multiple_instances.cc
VivekPanyam: This is not an issue here. The lambda captures idx by value so it is making a copy.

source/neuropod/tests/test_ope_multiple_instances.cc
VivekPanyam: This is heavily based off of the repro code you provided in #397. I missed the `unsigned long` when I was turning it into a test. Will update.

Don't think the explicit `size_t` is necessary.

source/neuropod/tests/test_ope_multiple_instances.cc
VivekPanyam: Adding it. Will retest to confirm it still catches the original issue

source/neuropod/multiprocess/shm_tensor.hh
```diff
@@ -192,9 +192,10 @@ public:
...
-        auto base_ptr =
-            stdx::make_unique<SHMNeuropodTensor<uint8_t>>(dims, std::move(block), data, block_id)->get_raw_data_ptr();
-        auto max_len = dims[dims.size() - 1];
+        auto byte_block = stdx::make_unique<SHMNeuropodTensor<uint8_t>>(dims, std::move(block), data, block_id);
+
+        auto base_ptr = byte_block->get_raw_data_ptr();
+        auto max_len  = dims[dims.size() - 1];
 
         // Copy into our local buffer
         auto numel = get_num_elements();
```
vkuzmin-uber, line 201: btw, just found, this is not related to you change but anyway:
1) It would be more consistent to init numel with write_buffer_.size() here (or just use write_buffer_.size() directly):
auto numel = write_buffer_.size().

2) I found that then for loop looks like this:
```
// Copy into our local buffer
        auto numel = get_num_elements();
        for (int i = 0; i < numel; i++)
        {
            auto pos     = base_ptr + i * max_len;
```
You see, numel is size_t here, iterator i is int. Theoretically this will cause infinite loop on 64 bit machine where int is 32bit and numel is greater than it can fit.

I saw some other places where for(int i =) was used, I am surprised that static check/linter doesn't report it. I understand that this is not critical on current scale but it would be nice to have it consistent with 64bit. In this case if we decide to test Neuropod's limits on a large numbers, we would not hit these obvious issues.

vkuzmin-uber (commented):
Left minor comment, not related to the fix but issues around the scope

vkuzmin-uber (approved):

VivekPanyam:
> You see, numel is size_t here, iterator i is int. Theoretically this will cause infinite loop on 64 bit machine where int is 32bit and numel is greater than i can accept w/o overflow.
>
> I saw some other places where for(int i =) was used, I am surprised that static check/linter doesn't report it. I understand that this is not critical on current scale but it would be nice to have it consistent with 64bit. In this case if we decide to test Neuropod's limits on a large numbers, we would not hit these obvious issues.


`clang-tidy` does find issues like this https://clang.llvm.org/extra/clang-tidy/checks/bugprone-too-small-loop-variable.html

but its default is to filter out cases where the "loop variable’s type has more magnitude bits as the specified upper limit. The default value is 16."

We can increase the limit once I get `clang-tidy` running in CI.

PR #435, opened 2020-09-17
[Misc] Fix many clang-tidy issues in preparation for enabling in CI

VivekPanyam:
### Summary:
This PR fixes many `clang-tidy` issues in preparation for enabling it in CI.

This PR should be a noop and all tests should pass.

### Test Plan:

CI


source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -167,7 +167,7 @@ void insert_value_in_output(NeuropodValueMap & output,
 #endif
 
         // if tensor_type string or no tensor_type and empty list or list containing actual string
-        if ((has_type && tensor_type == TensorType::STRING_TENSOR) || (!has_type && list.size() == 0) ||
+        if ((has_type && tensor_type == TensorType::STRING_TENSOR) || (!has_type && list.empty()) ||
```
vkuzmin-uber, line 170: Condition can be simplified to be more straightforward and exactly what comment says:

has_type && tensor_type == TensorType::STRING_TENSOR || !has_type && (list.empty() || list[0].isString())

source/neuropod/backends/torchscript/torch_backend.cc
```diff
@@ -433,6 +431,7 @@ std::unique_ptr<NeuropodValueMap> TorchNeuropodBackend::infer_internal(const Neu
         if (is_named_tuple)
         {
             // This is a named tuple
+            // NOLINTNEXTLINE(modernize-loop-convert): Can't always use a range based loop here
```
vkuzmin-uber, line 434: As you mentioned, with setting 31 instead of 16bit, we should ask clang-tidy to catch the problem like the one below https://clang.llvm.org/extra/clang-tidy/checks/bugprone-too-small-loop-variable.html

vkuzmin-uber (approved):
Left few minor comments, please look, just in case. Approved.

selitvin (approved):

source/neuropod/backends/torchscript/torch_backend.cc
VivekPanyam: I may enable that in a follow up PR, but the issue will likely be taken care of by some additional compile time warnings I'm enabling.

(For anyone reading, see my comment here for more details on this `clang-tidy` check: https://github.com/uber/neuropod/pull/433#issuecomment-693768830)

source/neuropod/backends/torchscript/torch_backend.cc
VivekPanyam: Probably not worth making any non `clang-tidy` related changes in this PR just to minimize the risk of introducing an issue (as this is a fairly large PR)

PR #436, opened 2020-09-17
Add a Torch 1.6.0 build to CI

VivekPanyam:
### Summary:
No major code changes were required to add support for Torch 1.6.0. This PR adds a Torch 1.6.0 build to CI.

A future PR will likely remove the Torch 1.1.0 build from CI in order to limit the resources we use on every PR.

### Test Plan:
- CI 
- Internal testing at ATG (where we've used Torch 1.6.0 with Neuropod for a while)


source/neuropod/internal/backend_registration.cc
```diff
@@ -63,12 +63,16 @@ const std::vector<BackendLoadSpec> default_backend_for_type = {
     {"torchscript", "1.2.0", "libneuropod_torchscript_backend.so"},
     {"torchscript", "1.3.0", "libneuropod_torchscript_backend.so"},
     {"torchscript", "1.4.0", "libneuropod_torchscript_backend.so"},
+    {"torchscript", "1.5.0", "libneuropod_torchscript_backend.so"},
```
vkuzmin-uber, line 66: I remember I was able to build our service with Torch 1.5 and tests passed successfully. I guess it because our model doesn't specify version but platform only and this is why correct backend was registered anyway.

vkuzmin-uber (approved):
I see you mentioned that need to remove 1.1.0, we may also need to upgrade default versions installed by script:

neuropod:> grep VERSION build/install_frameworks.py | grep getenv
REQUESTED_TF_VERSION = os.getenv("NEUROPOD_TENSORFLOW_VERSION") or "1.12.0"
REQUESTED_TORCH_VERSION = os.getenv("NEUROPOD_TORCH_VERSION") or "1.1.0"
CUDA_VERSION = os.getenv("NEUROPOD_CUDA_VERSION") or "10.0"

Also, we may consider not using default versions for build at all and by default build only explicitly selected backends. If TF and Torch versions are not set, build Python backend only (this is useful for projects that build neuropod from sources and often need only 1 backend)

selitvin (approved):

VivekPanyam:
Superseded by #456

PR #437, opened 2020-09-17
[Build] Start enabling additional warnings

VivekPanyam:
### Summary:
This PR introduces a `neuropod_cc_library` rule that builds targets with `-Weverything` (+ some exceptions).

All uses of `cc_library` within `source/internal` are modified to use `neuropod_cc_library` in this PR.

Once this PR lands, we can incrementally refactor the codebase to transition all first-party `cc_*` targets to use `neuropod_cc_*` rules. At that point we can build with `-Werror` in CI to ensure that newly introduced code doesn't add warnings to the build.

### Test Plan:

Local testing + CI

### DIFF
--- a/source/neuropod/internal/neuropod_tensor.cc
+++ b/source/neuropod/internal/neuropod_tensor.cc
@@ -33,13 +33,13 @@ std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims)
     std::vector<int64_t> out(dims.size());
 
     int64_t running_product = 1;
-    for (int i = out.size() - 1; i >= 0; i--)
+    for (size_t i = out.size(); i > 0; i--)
     {
         // Set the stride
-        out[i] = running_product;
+        out[i - 1] = running_product;
 
         // Update the running product
-        running_product *= dims[i];
+        running_product *= dims[i - 1];
     }
 
     return out;
@@ -48,7 +48,7 @@ std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims)
 size_t compute_num_elements(const std::vector<int64_t> &dims)
 {
     // Get the number of elements in the tensor by multiplying all the dims together
-    return std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>());
+    return static_cast<size_t>(std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>()));
 }
 
 } // namespace
@@ -116,7 +116,7 @@ bool NeuropodTensor::operator==(const NeuropodTensor &other) const
         const auto numel = get_num_elements();
         const auto t1    = as_typed_tensor<std::string>();
         const auto t2    = other.as_typed_tensor<std::string>();
-        for (int i = 0; i < numel; i++)
+        for (size_t i = 0; i < numel; i++)
         {
             if ((*t1)[i] != (*t2)[i])
             {
--- a/source/neuropod/internal/neuropod_tensor.hh
+++ b/source/neuropod/internal/neuropod_tensor.hh
@@ -155,7 +155,7 @@ public:
 
     NeuropodTensor(const NeuropodTensor &) = delete;
 
-    virtual ~NeuropodTensor() {}
+    virtual ~NeuropodTensor() override {}
 
     friend std::ostream &operator<<(std::ostream &out, const NeuropodTensor &tensor)
     {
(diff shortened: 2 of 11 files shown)


source/neuropod/internal/neuropod_tensor.cc
```diff
@@ -48,7 +48,7 @@ std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims)
 size_t compute_num_elements(const std::vector<int64_t> &dims)
 {
     // Get the number of elements in the tensor by multiplying all the dims together
-    return std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>());
+    return static_cast<size_t>(std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>()));
```
vkuzmin-uber, line 51: I remember we had negative dims (-2 for symbol shapes). At this level, is it possible that we have valid negative dim?

source/neuropod/internal/neuropod_tensor.cc
```diff
@@ -33,13 +33,13 @@ std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims)
     std::vector<int64_t> out(dims.size());
 
     int64_t running_product = 1;
-    for (int i = out.size() - 1; i >= 0; i--)
+    for (size_t i = out.size(); i > 0; i--)
```
vkuzmin-uber, line 36: interesting way to avoid overflow. I usually used either ssize_t (this is from C10 I guess) or long long in for loop and use the same approach with negative i at the end.

source/neuropod/internal/neuropod_tensor.hh
```diff
@@ -155,7 +155,7 @@ public:
 
     NeuropodTensor(const NeuropodTensor &) = delete;
 
-    virtual ~NeuropodTensor() {}
+    virtual ~NeuropodTensor() override {}
```
vkuzmin-uber, line 158: If this is override, it means its Base class already has virtual dtor. In this case you may just remove it completely (as I remember this considered as a best practice that Derived class doesn't specify empty virtual dtor if its Base has virtual dtor already)

vkuzmin-uber (approved):
Some comments

source/neuropod/internal/neuropod_tensor.hh
vkuzmin-uber: unrelated, just reminded me some idea I had once and I want to share with you:

you know that STL containers doesn't have virtual dtor (because "virtual" is not free and STL supports algos then classic OOP approach). But because of that "public inheritance" from STL containers introduces problem of resource release in case of "delete pBase".

I understand that it will never happen with current STL but still theoretically it could solve this issue by allowing user to parametrize explicitly "virtual dtor"/"regular dtor":

STL would need to add one more template param:
 
template<..., class BasewithDtor=BaseWithRegularDtor>
class vector: public BaseWithDtor {...}

struct BaseWithRegularDtor{ virtual ~BaseWithRegularDtor() {} }
struct BaseWithVirtualDtor{ virtual ~BaseWithVirtualDtor() {} }

class MyContainer: public std::vector<..., BasewithDtor=BaseWithVirtualDtor> {}

I never used this technic in real project, this design patter makes sense only for libs like STL  

What do you think, would this "design pattern" can be useful theoretically ?

dapirian (approved):
looks familiar 😉

source/neuropod/internal/neuropod_tensor.cc
VivekPanyam: `compute_num_elements` is used on tensors (not tensor specs) so it is not valid to have negative dims at this point

source/neuropod/internal/neuropod_tensor.cc
VivekPanyam: If I used a signed type (like `int` or `ssize_t`), I'd need to do static casts within the loop to avoid additional warnings. I'm okay with either solution.

Do you think the static cast solution is more clear?

source/neuropod/internal/neuropod_tensor.hh
VivekPanyam: Maybe there are situations where it would be useful, but, with the design above, you lose a lot of the benefits of inheritance since `std::vector<..., VirtualDtor>` and `std::vector<...>` are not interchangeable (in terms of dynamic polymorphism). The remaining major benefit I see would be code reuse, but composition would likely be a better solution than inheritance here.

Also, in your implementation above, both `RegularDtor` and `VirtualDtor` define virtual destructors. A "real" implementation of this likely won't be straightforward.

Either way, I don't think it would be useful here because `NeuropodTensor` is an abstract class and therefore doesn't have any "non-virtual" version

PR #438, opened 2020-09-19
[Build] Refactor tests and coverage collection

VivekPanyam:
### Summary:
This PR refactors build and test scripts to use Bazel to directly run tests. It also lets us collect Java coverage information in a more efficient way than was being done previously.

As part of this, we upgrade Bazel to 3.1.0 and also make the build more isolated from the environment (by making some changes in `python.bzl` and in `.bazelrc`).

### Test Plan:

CI + manual validation of the coverage report


VivekPanyam:
Looking at the coverage report, it seems like the new way of collecting C++ coverage (converting the C++ report to `lcov` before uploading to codecov) includes lines that are just a closing brace (`}`) as executable lines.

The previous method did not count these lines as executable.

This is adding 817 lines to the report (out of which 663 are considered covered). Practically, this doesn't mean anything, but does drop the overall coverage numbers by ~1%

source/.bazelrc
```diff
@@ -21,10 +21,22 @@ build --cxxopt='-D_GLIBCXX_USE_CXX11_ABI=0'
 build --extra_toolchains=@bazel_tools//tools/python:autodetecting_toolchain_nonstrict
 
 # Build with coverage info
-build:coverage --copt='-fprofile-instr-generate' --copt='-fcoverage-mapping' --linkopt='-fprofile-instr-generate' --linkopt='-fcoverage-mapping'
+build:coverage --copt='-fprofile-instr-generate' --copt='-fcoverage-mapping' --linkopt='-fprofile-instr-generate' --linkopt='-fcoverage-mapping' --run_under=//tools/coverage:run_under
+
+# Java coverage settings
+build:coverage --collect_code_coverage --instrumentation_filter='//neuropod/bindings/java:neuropod_java(_jar)?'
```
vkuzmin-uber, line 27: Interesting, it doesn't need:

coverage_report_generator=@bazel_tools//tools/test/CoverageOutputGenerator/java/com/google/devtools/coverageoutputgenerator:Main

I guess this is because of upraded bazel. Anyway, if it works, this is great that this is optional now.

vkuzmin-uber (approved):
LGTM

PR #440, opened 2020-09-21
[Build] Treat build warnings as errors in CI

VivekPanyam:
### Summary:

Build with `-Werror` in CI so build warnings are treated as errors

### Test Plan:

CI


vkuzmin-uber (approved):
I think this is a right move.

PR #441, opened 2020-09-22
Add a `view` and `flat` method to TypedNeuropodTensor

VivekPanyam:
### Summary:

This PR adds a `view` method to `TypedNeuropodTensor<T>` that allows users to view and access a tensor as if it had a different set of dimensions.

Somewhat similar to https://pytorch.org/docs/stable/tensors.html#torch.Tensor.view except we currently don't support `-1` dimensions. Our implementation is also simpler because all `NeuropodTensor`s are currently contiguous.

The `view` method returns a `TensorView` which is a thin wrapper around `TensorAccessor` to store strides and dimensions.

Changes made in the returned view modify the original tensor data.

`flat` is a simple method that returns a 1D view of the tensor (effectively `return view(get_num_elements())`). See https://www.tensorflow.org/api_docs/cc/class/tensorflow/tensor#flat for more details.

### Test Plan:

CI + added tests

### DIFF
--- a/source/neuropod/internal/neuropod_tensor.cc
+++ b/source/neuropod/internal/neuropod_tensor.cc
@@ -17,10 +17,23 @@ limitations under the License.
 
 #include "neuropod/internal/error_utils.hh"
 
+#include <algorithm>
+
 namespace neuropod
 {
 
 namespace
+{
+
+size_t compute_num_elements(const std::vector<int64_t> &dims)
+{
+    // Get the number of elements in the tensor by multiplying all the dims together
+    return std::accumulate(dims.begin(), dims.end(), static_cast<size_t>(1), std::multiplies<>());
+}
+
+} // namespace
+
+namespace detail
 {
 
 std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims)
@@ -45,17 +58,6 @@ std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims)
     return out;
 }
 
-size_t compute_num_elements(const std::vector<int64_t> &dims)
-{
-    // Get the number of elements in the tensor by multiplying all the dims together
-    return static_cast<size_t>(std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>()));
-}
-
-} // namespace
-
-namespace detail
-{
-
 void throw_error_hh(const char *file, int line, const char *function, const std::string &message, TensorType type)
 {
     throw_error(file, line, function, message, type);
@@ -73,7 +75,7 @@ NeuropodTensor::NeuropodTensor(TensorType tensor_type, const std::vector<int64_t
     : NeuropodValue(true),
       tensor_type_(tensor_type),
       dims_(dims),
-      strides_(compute_strides(dims)),
+      strides_(detail::compute_strides(dims)),
       num_elements_(compute_num_elements(dims)),
       device_(device)
 {
@@ -148,6 +150,27 @@ void NeuropodTensor::assure_device_cpu() const
     {
         NEUROPOD_ERROR("Tried to perform an operation on a tensor that expected the tensor to be on CPU. Tensor: {}",
                        *this);
+    }
+}
+
+void NeuropodTensor::assure_view_compatible_shape(const std::vector<int64_t> &requested_dims) const
+{
+    // Make sure all provided dimensions are positive
+    if (!std::all_of(requested_dims.begin(), requested_dims.end(), [](int64_t i) { return i > 0; }))
+    {
+        NEUROPOD_ERROR("All dimensions provided to `view` must be positive");
+    }
+
+    // Compute the total number of elements
+    auto new_num_elements = compute_num_elements(requested_dims);
+    auto num_elements     = get_num_elements();
+
+    if (new_num_elements != num_elements)
+    {
+        NEUROPOD_ERROR("The requested view dimensions are not compatible with the tensor. Requested num "
+                       "elements: {}. Actual num elements: {}",
+                       new_num_elements,
+                       num_elements);
     }
 }
 
--- a/source/neuropod/internal/neuropod_tensor.hh
+++ b/source/neuropod/internal/neuropod_tensor.hh
@@ -69,6 +69,9 @@ namespace detail
 [[noreturn]] void throw_error_hh(
     const char *file, int line, const char *function, const std::string &message, TensorType type1, TensorType type2);
 
+// Utility to compute strides given a vector of dimensions
+std::vector<int64_t> compute_strides(const std::vector<int64_t> &dims);
+
 } // namespace detail
 
 // Forward declare NeuropodTensor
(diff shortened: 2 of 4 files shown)


source/neuropod/internal/neuropod_tensor.cc
```diff
@@ -23,6 +23,17 @@ namespace neuropod
 namespace
 {
 
+size_t compute_num_elements(const std::vector<int64_t> &dims)
+{
+    // Get the number of elements in the tensor by multiplying all the dims together
+    return static_cast<size_t>(std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<>()));
```
vkuzmin-uber, line 29: #include <functional>   // std:: multiplies
#include <numeric>      // std::accumulate

source/neuropod/internal/neuropod_tensor.cc
```diff
@@ -151,6 +151,27 @@ void NeuropodTensor::assure_device_cpu() const
     }
 }
 
+void NeuropodTensor::assure_view_compatible_shape(const std::vector<int64_t> &requested_dims) const
+{
+    // Make sure all provided dimensions are positive
+    if (!std::all_of(requested_dims.begin(), requested_dims.end(), [](int64_t i) { return i > 0; }))
```
vkuzmin-uber, line 157: #include <algorithm>

source/neuropod/internal/neuropod_tensor.cc
vkuzmin-uber: btw, it may overflow theoretically but maybe at this moment we may assume that dims are valid for this operation

source/neuropod/internal/neuropod_tensor.hh
```diff
@@ -460,7 +514,9 @@ public:
     static int cmp(const std::string &lhs, const std::string &rhs) { return lhs.compare(rhs); }
 
     // Based on https://en.cppreference.com/w/cpp/language/operators
+    // TODO(vip): Simplify these operators
```
vkuzmin-uber, line 517: Maybe in C++ 20 https://devblogs.microsoft.com/cppblog/simplify-your-code-with-rocket-science-c20s-spaceship-operator/

source/neuropod/tests/test_accessor.cc
```diff
@@ -166,4 +166,10 @@ TEST(test_accessor, test_string_write)
         auto actual = tensor->get_data_as_vector();
         EXPECT_EQ(expected, actual);
     }
+
+    // Get a flat view
+    {
+        auto actual = tensor->flat();
```
vkuzmin-uber, line 172: Test for operator[] would be nice, for read and write both

vkuzmin-uber (commented):
Left comment, test for operator [], read/write.

source/neuropod/internal/neuropod_tensor.hh
```diff
@@ -338,6 +344,54 @@ public:
...
+        assure_view_compatible_shape(new_dims);
+
+        // Create and return a tensor view
+        constexpr size_t num_dims    = sizeof...(requested_dims);
+        auto             new_strides = detail::compute_strides(new_dims);
+        return TensorView<T *, num_dims>(get_raw_data_ptr(), std::move(new_dims), std::move(new_strides));
+    }
+
+    template <typename... Dim>
+    auto view(Dim... requested_dims) const
```
vkuzmin-uber, line 367: I see const/non-const are different in last line, T/"const T'. I think it won't be hard to move it to private helper template method "view_type" that accepts type and then can be used by both views.

source/neuropod/internal/neuropod_tensor.cc
```diff
@@ -151,6 +151,27 @@ void NeuropodTensor::assure_device_cpu() const
...
+    {
+        NEUROPOD_ERROR("All dimensions provided to `view` must be positive");
+    }
+
+    // Compute the total number of elements
+    auto new_num_elements = compute_num_elements(requested_dims);
+    auto num_elements     = get_num_elements();
+
+    if (new_num_elements != num_elements)
```
vkuzmin-uber, line 166: If we so strictly check it, we do need requested_dims at all?

Do we expect that requested_dims can be ever different than Tensor's dims?

vkuzmin-uber (commented):
Added few more comments

source/neuropod/internal/neuropod_tensor.cc
VivekPanyam: We're checking that the number of elements match, not that the dims match.

In fact, we expect that `requested_dims` will almost always be different than the tensor's dims.

The purpose of `view` is to view a tensor as if it had a different set of dimensions. The major restriction here is that the total number of elements must be the same. 

For example, viewing a `(10, 10, 3)` tensor as a `(10, 30)` or `(1, 300)` or `(300,)` tensor is fine because the total number of elements is the same.

`view` is quite similar to `reshape` in numpy except it guarantees that a copy of the underlying data is not made (see https://numpy.org/doc/stable/reference/generated/numpy.reshape.html for more details). Also see the link to `view` in pytorch in the PR description

vkuzmin-uber: I see, I expected that having required_dims should allow even if we have "smaller" window but I guess requirement for same "size" is fine.

source/neuropod/tests/test_accessor.cc
```diff
@@ -166,4 +166,51 @@ TEST(test_accessor, test_string_write)
...
+        {
+            for (int j = 0; j < 5; j++)
+            {
+                view[i][j] = i * 5 + j;
+            }
+        }
+    }
+
+    std::vector<int32_t> expected(15);
+    std::iota(expected.begin(), expected.end(), 0);
```
vkuzmin-uber, line 213: This is cool, first time I see it, somehow missed this feature of C++11.

btw, Go has "iota" for emum/const initialization https://yourbasic.org/golang/iota/

source/neuropod/internal/neuropod_tensor.cc
```diff
@@ -17,12 +17,25 @@ limitations under the License.
...
 namespace neuropod
 {
 
 namespace
 {
 
+size_t compute_num_elements(const std::vector<int64_t> &dims)
+{
+    // Get the number of elements in the tensor by multiplying all the dims together
+    return std::accumulate(dims.begin(), dims.end(), static_cast<size_t>(1), std::multiplies<>());
```
vkuzmin-uber, line 31: Not critical, I left comments in prev iteration, exact headers for these algos are:

#include <functional> // std:: multiplies
#include <numeric> // std::accumulate

vkuzmin-uber (approved):
LGTM, left minor comment about "include headers"

source/neuropod/internal/neuropod_tensor.cc
VivekPanyam: Ah I forgot to submit my response to your comments. Based on [`include-what-you-use`](https://include-what-you-use.org/) from Google:
> "Include what you use" means this: for every symbol (type, function variable, or macro) that you use in foo.cc, either foo.cc or foo.h should #include a .h file that exports the declaration of that symbol. 

For `std::accumulate`, my updates did add `#include <algorithm>` to the `.cc` file

For `std::multiplies`, `neuropod_tensor.hh` includes `<functional>`

VivekPanyam: _Note: after looking just now, the include-what-you-use definition contradicts the [Google style guide recommendation](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use):_

> ... foo.cc should include bar.h if it uses a symbol from it even if foo.h includes bar.h. ...

That said, I think it's okay for us to stick to the definition in my comment above. If we introduce IWYU into lint/CI, we can enforce it throughout the codebase.

VivekPanyam: Tracked by #446

vkuzmin-uber: Just found that github formatter hided header name because of not escaped \<. Just minor note:

https://en.cppreference.com/w/cpp/algorithm/accumulate "Defined in header <numeric>" , I guess that "algorithm" includes it as well and this is why it works and is under the same "include-what-you-use" topic you mentioned.

Thanks for creating Issue.

PR #442, opened 2020-09-22
[Build] Refactor backends to adhere to stricter build warnings

VivekPanyam:
### Summary:

Refactor code under `source/neuropod/backends` to adhere to the stricter set of warnings introduced in #437

### Test Plan:

CI

### DIFF
--- a/source/neuropod/backends/torchscript/torch_tensor.hh
+++ b/source/neuropod/backends/torchscript/torch_tensor.hh
@@ -29,6 +29,11 @@ limitations under the License.
 // If we're not building with a nightly relase of torch,
 // set the date to match the date of the official release
 #ifndef CAFFE2_NIGHTLY_VERSION
+#if CAFFE2_VERSION == 10100
+// The date of the official torch 1.1.0 release
+#define CAFFE2_NIGHTLY_VERSION 20190430
+#endif
+
 #if CAFFE2_VERSION == 10200
 // The date of the official torch 1.2.0 release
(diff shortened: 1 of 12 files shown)


source/neuropod/backends/torchscript/torch_tensor.hh
```diff
@@ -254,9 +259,6 @@ protected:
 };
 
 // Utility function to get an IValue from a torch tensor
-torch::jit::IValue get_ivalue_from_torch_tensor(const std::shared_ptr<NeuropodValue> &tensor)
-{
-    return std::dynamic_pointer_cast<NativeDataContainer<torch::jit::IValue>>(tensor)->get_native_data();
-}
+torch::jit::IValue get_ivalue_from_torch_tensor(const std::shared_ptr<NeuropodValue> &tensor);
```
vkuzmin-uber, line 262: you could make it "inline", this is legal way to provide definition in header w/o confusing compiler

vkuzmin-uber (approved):
LGTM

source/neuropod/backends/torchscript/torch_tensor.hh
VivekPanyam: Yup! Previously, I marked it as `inline` in an internal patch on top of Neuropod, but decided to pull it into a `.cc` file here.

I was planning on pulling more things into the `.cc` file, but taking another look, it probably makes sense to just `inline` this in the header and leave it alone

sepo-eng (approved):

PR #443, opened 2020-09-22
[Build] Refactor OPE code to adhere to stricter build warnings

VivekPanyam:
### Summary:

Refactor code under `source/neuropod/multiprocess` to adhere to the stricter set of warnings introduced in #437

### Test Plan:

CI


source/neuropod/multiprocess/multiprocess.cc
```diff
@@ -76,17 +76,18 @@ pid_t start_worker_process(const std::string &control_queue_name, std::vector<st
     // Setup the environment
 
     // Null terminated char * array
-    char *env_arr[env.size() + 1];
+    std::vector<char *> env_arr(env.size() + 1);
```
vkuzmin-uber, line 79: in case of vector, all elements are initialized with 0 (nullptr) already because it calls (char*) init for every element, so next line is redundant (it was necessary for regular array though)

source/neuropod/multiprocess/shm_tensor.hh
```diff
@@ -186,20 +186,20 @@ public:
...
+                                       std::shared_ptr<void>       block,
+                                       shm_tensor *                data,
+                                       const SHMBlockID &          block_id)
         : TypedNeuropodTensor<std::string>(copy_and_strip_last_dim(dims)), write_buffer_(this->get_num_elements())
     {
         auto byte_block = stdx::make_unique<SHMNeuropodTensor<uint8_t>>(dims, std::move(block), data, block_id);
 
         auto base_ptr = byte_block->get_raw_data_ptr();
-        auto max_len  = dims[dims.size() - 1];
+        auto max_len  = static_cast<size_t>(dims[dims.size() - 1]);
```
vkuzmin-uber, line 198: I think this cast is not needed and as result you don't need to cast it then at line 229 again

source/neuropod/multiprocess/tensor_utils.hh
```diff
@@ -27,16 +27,16 @@ namespace neuropod
 // This is useful for serialization and to wrap and/or copy tensors between backends.
 // For example, if you had a TorchNeuropodTensor and you wanted to get a tensor compatible
 // with `allocator` without making a copy, you could use this function
-std::shared_ptr<NeuropodTensor> wrap_existing_tensor(NeuropodTensorAllocator &       allocator,
-                                                     std::shared_ptr<NeuropodTensor> tensor)
+static std::shared_ptr<NeuropodTensor> wrap_existing_tensor(NeuropodTensorAllocator &       allocator,
```
vkuzmin-uber, line 30: Interesting, did compiler require it to be static? In C++, the same can be achieved with anonymous namespace

namespace
{

// functions here
}

vkuzmin-uber (approved):
Looks good. Please check I left comment and question.

sepo-eng (approved):

PR #444, opened 2020-09-22
[Build] Refactor more code to adhere to stricter build warnings

VivekPanyam:
### Summary:

Refactor additional code (mostly under `source/neuropod/bindings`) to adhere to the stricter set of warnings introduced in #437

### Test Plan:

CI


source/neuropod/tests/ope_overrides.hh
```diff
@@ -25,7 +25,7 @@ limitations under the License.
 namespace detail
 {
 
-std::string get_cwd()
+static std::string get_cwd()
```
vkuzmin-uber, line 28: asked already about static in other review

source/neuropod/serialization/serialization.cc
```diff
@@ -88,7 +88,7 @@ std::shared_ptr<NeuropodValue> deserialize(boost::archive::binary_iarchive &ar,
 
 void serialize(boost::archive::binary_oarchive &out, const NeuropodValueMap &item)
 {
-    int num_items = item.size();
+    auto num_items = static_cast<int>(item.size());
```
vkuzmin-uber, line 91: is it supposed to be int here?

vkuzmin-uber (approved):
Looks good. One comment about serialization, please check

source/neuropod/tests/ope_overrides.hh
vkuzmin-uber: maybe anonymous namespace instead of detail will work here instead of "static"

sepo-eng (approved):

Issue #348, opened 2020-05-14
Allow models to specify target framework versions


VivekPanyam:
At export time, models should be able to specify version(s) of a framework that the model is compatible with.

When we load a model, Neuropod will attempt to find a backend that satisfies the specified range.

This combined with OPE will enable environments to provide many supported versions of each framework and have models choose which one to use.

This is useful in several cases:
- If a model uses custom ops that are only compatible with one version of the framework (e.g. custom ops built against TF 1.15 headers vs TF 1.13.1 headers)
- Being able to ensure that runtime and test environments are as similar as possible
- Avoiding bugs in a specific version of the underlying framework

Target version ranges are specified as semver ranges (see https://semver.org/, https://docs.npmjs.com/misc/semver#ranges, https://docs.npmjs.com/misc/semver#advanced-range-syntax) which provides a very flexible way of defining the required framework. 

## TODO:

- [x] Specify target versions at export time (#316)
- [x] Load different backends depending on the specified version range (#330)
- [x] Better error messages about required framework versions when loading models from python (#408)
- [ ] Rename backend `so` files depending on the version of the framework they're built against (e.g. `libneuropod_torchscript_1_4_0_cpu_backend.so`)

Once this is production ready, using dockerized OPE worker processes will also let us specify required CUDA versions (if any) and provide even more isolation.

PR #447, opened 2020-10-06
Refactor backend loading to also check absolute paths

VivekPanyam:
### Summary:
This PR adds some comments to the backend loading system and refactors the loading behavior to support default backends installed at absolute paths.

This helps us start to solve a few problems:
- There are a growing number of interfaces to Neuropod (C++, Python, C, Java, etc) and we don't want unique ways of installing backends per language.
- There are also issues with Java and `LD_LIBRARY_PATH` on macOS because of system integrity protection. This makes it quite difficult to use `LD_LIBRARY_PATH` to add backends for Neuropod to dlopen
-  Models can't choose the version of a framework to use. This is because all the backend `so` files for a type have the same name regardless of the version (so the backend loading system cannot load a specific version). Even if they had different names based on version (e.g. `libneuropod_torchscript_1_4_0_cpu_backend.so`), we would still see issues when adding their containing folders to the library loader search path. This is because then there would be conflicting versions of other unversioned shared objects included with the backends (e.g. `libc10.so` from torch 1.1.0 and `libc10.so` from torch 1.4.0)

We start to solve these by allowing users to install backends at a "well known" path on the system. This way, all installed backends will be available from all languages without any additional configuration.

For now, the "well known" location is `/usr/local/lib/neuropod/`, but this may change or become user-configurable in the future.

We solve the last problem mentioned above by installing each backend in a versioned location (that is not on the system's library path). For example, `/usr/local/lib/neuropod/0.2.0/backends/torchscript_1.4.0_gpu/libneuropod_torchscript_backend.so`.

This means each backend version can now be independently loaded without being on the system library path. See https://github.com/uber/neuropod/issues/348 for more details. 

### Test Plan:
CI + Local testing

Will explore tests with multiple framework versions in CI

### DIFF
--- a/source/neuropod/internal/backend_registration.cc
+++ b/source/neuropod/internal/backend_registration.cc
@@ -16,12 +16,14 @@ limitations under the License.
 #include "backend_registration.hh"
 
 #include "neuropod/internal/error_utils.hh"
+#include "neuropod/version.hh"
 
 #include <iostream>
 #include <mutex>
 #include <sstream>
 #include <stdexcept>
 #include <unordered_map>
+#include <unordered_set>
 
 #include <cpp-semver.hpp>
 #include <dlfcn.h>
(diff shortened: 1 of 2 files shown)


sepo-eng (approved):
LGTM.

The get_default_backend_map seems like a good candidate for some unit tests.

source/neuropod/internal/backend_registration.cc
```diff
@@ -51,44 +53,74 @@ void init_registrar_if_needed()
...
+
+                for (const auto &version : framework.versions)
+                {
+                    BackendLoadSpec item;
+                    item.type    = framework.type;
+                    item.version = version;
+
+                    if (is_absolute_path)
+                    {
+                        // Ex:
```
sepo-eng, line 99: Is it worth validating the existence of the file if is_absolute_path is true, and skip appending to 'out' if we find it doesn't exist?

VivekPanyam: Probably not for the following reasons:
- If a file exists, it doesn't necessarily mean `dlopen` will work
- For relative paths, internally `dlopen` effectively just tries a bunch of absolute paths based on the system loader paths and `LD_LIBRARY_PATH`. Therefore, the approach here of trying several candidate locations likely isn't going to be "expensive"

VivekPanyam:
> The get_default_backend_map seems like a good candidate for some unit tests.

I have a few PRs in progress related to library loading + changes in CI to make dynamic loading tests easier to build.

Not sure how much value we get out of testing `get_default_backend_map` directly in its current form. Its purpose was to avoid writing out the (fairly large) backend list manually in the code. We could spot check the output of the function, but I'd rather have "exhaustive" multi-version library loading tests in CI to ensure we can actually load all of these backends correctly.

Maybe it makes sense to refactor `get_default_backend_map` to take in a `std::vector<FrameworkInfo>` instead of using a fixed hardcoded one

PR #448, opened 2020-10-06
[Python] Refactor python interpreter state management

VivekPanyam:
### Summary:

(will add before requesting review)

### DIFF
--- a/source/neuropod/backends/python_bridge/python_bridge.cc
+++ b/source/neuropod/backends/python_bridge/python_bridge.cc
@@ -23,8 +23,6 @@ limitations under the License.
 #include <sstream>
 #include <vector>
 
-#include <dlfcn.h>
-
 namespace neuropod
 {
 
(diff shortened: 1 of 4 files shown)


VivekPanyam:
`Py_Finalize` calls can create lots of tricky issues because of C++ and python object lifetimes. For this reason, Boost.Python doesn't support finalizing interpreters.

pybind11 is better about this, but the user is responsible for correctly managing object lifetimes. Unfortunately, even if we do everything correctly, incorrect cleanup in third party python bindings (e.g. torch) can cause crashes.

An example related to torch specifically: https://github.com/pytorch/pytorch/issues/23540

One issue with not calling `Py_Finalize` is that python `atexit` handlers never get called. This is important for code coverage in CI (because coverage.py uses `atexit`). 

Closing for now because we currently only depend on `atexit` for coverage and there are cleaner ways to fix that usecase.

(If anyone has any good ideas for how to solve this in a clean way, feel free to comment) 

_Note: the solution implemented by the author of the linked issue doesn't work for us because we can't guarantee that there are no tensors laying around in static objects_

PR #451, opened 2020-10-14
Refactor backend loading to simplify installation

VivekPanyam:
### Summary:

This PR significantly simplifies packaging, installing, and loading backends from different languages.

Backends can now be "installed" in a well known location. After a backend is installed on your system, it should be possible to use it from any Neuropod interface (Python, C++, Java, Go, etc) with no additional configuration.

This change builds on top of #447 so look there first for more context. 

#447 introduced a well-known location that defaults to `/usr/local/lib/neuropod`. This PR makes it possible to override that path by setting the `NEUROPOD_BASE_DIR` environment variable.

Backends are installed by untarring the backend into that location (e.g. `tar -xf ./some_backend.tar.gz -C $NEUROPOD_BASE_DIR`). Each backend tar file contains the appropriate folder structure including the neuropod version, backend version, and device variant.

For example, the `tar` file for a TorchScript backend for Neuropod 0.2.0 and Torch 1.4.0 with GPU support creates the following folder structure: `/${NEUROPOD_BASE_DIR}/0.2.0/backends/torchscript_1.4.0_gpu/`

General notes:
- This does not break existing loading behavior as `default_backend_overrides`, `LD_LIBRARY_PATH`, etc will still work. See #447 for more details
- Once this PR lands, models that are running using OPE should be able to specify a required framework version and Neuropod should transparently load it (assuming a version that satisfies the requirement is installed). This is not tested in this PR as it'll require some significant modifications to CI pipelines, but will be tested in a future PR.
- Python wheels for backends are removed as they are no longer necessary

### Test Plan:

CI


source/neuropod/internal/backend_registration.cc
```diff
@@ -70,7 +70,14 @@ std::vector<BackendLoadSpec> get_default_backend_map()
     const std::vector<FrameworkInfo> frameworks = {
         {"torchscript", "libneuropod_torchscript_backend.so", true, {"1.1.0", "1.2.0", "1.3.0", "1.4.0", "1.5.0"}},
         {"tensorflow", "libneuropod_tensorflow_backend.so", true, {"1.12.0", "1.13.1", "1.14.0", "1.15.0"}},
-        {"python", "libneuropod_pythonbridge_backend.so", false, {"27", "35", "36", "37", "38"}}};
+        {"python", "libneuropod_pythonbridge_backend.so", false, {"2.7", "3.5", "3.6", "3.7", "3.8"}}};
+
+    // Base directory for Neuropod backends
+    std::string neuropod_base_dir = "/usr/local/lib/neuropod";
+    if (auto base_dir = std::getenv("NEUROPOD_BASE_DIR"))
```
vkuzmin-uber, line 77: This is example of whenn "Elvis operator" would be useful:

std::string neuropod_base_dir = std::getenv("NEUROPOD_BASE_DIR") ?: ""/usr/local/lib/neuropod""

vkuzmin-uber (approved):

source/neuropod/internal/backend_registration.cc
VivekPanyam: Yeah, unfortunately it isn't in the C++ standard - just a GCC extension

For anyone who's curious, there's a talk from CppCon2019 that goes over this: ["Non-conforming C++: the Secrets the Committee Is Hiding From You"](https://youtu.be/IAdLwUXRUvg?t=364) (linked to the relevant timestamp in the talk)

PR #452, opened 2020-10-21
[Python Bindings] Use a consistent string type regardless of Python version

VivekPanyam:
### Summary:
The previous logic for string tensors in the python bindings would return unicode arrays for Python 3 and string arrays for Python 2. This involved an implicit decode step which we don't want (in case the string data isn't a valid UTF8 string).

We always want to return arrays of dtype `np.string_` regardless of version.

The previous implementation used `py::array` and `py::cast` to do the conversion in one line, but this doesn't let us explicitly specify the type of the resulting array. The new implementation explicitly creates a string array and copies the tensor data in. This should have the same efficiency as the previous implementation, but just makes the copies explicit.

**Note: this is a breaking change for python3 models that take string tensors as inputs. The data type of the inputs will change from `np.unicode_` to `np.bytes_`**

### Test Plan:

CI + modified tests

### DIFF
--- a/source/neuropod/tests/test_data/pytorch_strings_model/0/code/strings_model.py
+++ b/source/neuropod/tests/test_data/pytorch_strings_model/0/code/strings_model.py
@@ -5,7 +5,19 @@ import torch.nn as nn
 
 class StringsModel(nn.Module):
     def forward(self, x, y):
-        return {"out": np.array([f + " " + s for f, s in zip(x, y)])}
+        # Python 3 uses unicode for strings. Neuropod doesn't automatically convert
+        # string tensors to unicode because it's possible that these tensors contain
+        # aribtrary byte sequences instead of valid unicode.
+
+        # Since we know x and y contain valid strings, we can convert to np.str_ here.
+        # This decodes the data as unicode in python3 and doesn't have an effect in
+        # python2.
+        x = x.astype(np.str_)
+        y = y.astype(np.str_)
+
+        return {
+            "out": np.array([f + " " + s for f, s in zip(x, y)])
+        }
 
 
 def get_model(_):
--- a/source/python/neuropod/backends/neuropod_executor.py
+++ b/source/python/neuropod/backends/neuropod_executor.py
@@ -175,7 +175,7 @@ class NeuropodExecutor(object):
         # Convert unicode to string
         for k, v in inputs.items():
             if v.dtype.type == np.unicode_:
-                inputs[k] = v.astype("str")
+                inputs[k] = v.astype(np.string_)
 
         # Validate inputs
         validate_tensors_against_specs(inputs, self.neuropod_config["input_spec"])
@@ -189,7 +189,7 @@ class NeuropodExecutor(object):
         # Convert unicode to string
         for k, v in out.items():
             if v.dtype.type == np.unicode_:
-                out[k] = v.astype("str")
+                out[k] = v.astype(np.string_)
 
         # Validate outputs
         validate_tensors_against_specs(out, self.neuropod_config["output_spec"])
--- a/source/python/neuropod/tests/test_pytorch_strings.py
+++ b/source/python/neuropod/tests/test_pytorch_strings.py
@@ -26,6 +26,16 @@ import torch.nn as nn
 
 class StringsModel(nn.Module):
     def forward(self, x, y):
+        # Python 3 uses unicode for strings. Neuropod doesn't automatically convert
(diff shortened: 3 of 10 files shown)


source/neuropod/tests/test_data/pytorch_strings_model/0/code/strings_model.py
```diff
@@ -5,7 +5,19 @@
 
 class StringsModel(nn.Module):
     def forward(self, x, y):
-        return {"out": np.array([f + " " + s for f, s in zip(x, y)])}
+        # Python 3 uses unicode for strings. Neuropod doesn't automatically convert
+        # string tensors to unicode because it's possible that these tensors contain
+        # aribtrary byte sequences instead of valid unicode.
```
vkuzmin-uber, line 10: typo: aribtrary

source/python/neuropod/tests/test_pytorch_strings.py
```diff
@@ -26,6 +26,16 @@
 
 class StringsModel(nn.Module):
     def forward(self, x, y):
+        # Python 3 uses unicode for strings. Neuropod doesn't automatically convert
+        # string tensors to unicode because it's possible that these tensors contain
+        # aribtrary byte sequences instead of valid unicode.
```
vkuzmin-uber, line 31: typo: aribtrary

source/python/neuropod/utils/dtype_utils.py
```diff
@@ -21,14 +21,14 @@ def get_dtype(arg):
...
-        arg = "str"
+        arg = "S"
 
     return np.dtype(arg)
 
 
 def get_dtype_name(arg):
     name = get_dtype(arg).name
-    if name == "str":
+    if name == "bytes":
```
vkuzmin-uber, line 31: "str" won't be converted to "string" and stay "star"
What is the effect of this change, can you leave comment?

source/python/neuropod/utils/eval_utils.py
```diff
@@ -29,9 +29,9 @@
 
 def check_output_matches_expected(out, expected_out):
     for key, value in expected_out.items():
-        if value.dtype.type == np.str_:
+        if value.dtype.type == np.string_ or value.dtype.type == np.str_:
             # All strings are equal
-            success_condition = (value == out[key]).all()
+            success_condition = (value.astype(np.string_) == out[key]).all()
```
vkuzmin-uber, line 34: For Python 3, if value is np.str_ that is unicode string and may have symbols that can't be converted with astype(nnp.string_).

Just checked:
python3
```
>>> b = np.array("привет", dtype=np.str_)
>>> b
array('привет', dtype='<U6')
>>> b.astype(np.string_)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-5: ordinal not in range(128)
```

So, covertion from np.string_ to np.str_ would be allowed always. 

Do you expect that conversion is always possible?

source/neuropod/tests/test_data/pytorch_strings_model/0/code/strings_model.py
```diff
@@ -5,7 +5,19 @@
 
 class StringsModel(nn.Module):
     def forward(self, x, y):
-        return {"out": np.array([f + " " + s for f, s in zip(x, y)])}
+        # Python 3 uses unicode for strings. Neuropod doesn't automatically convert
```
selitvin, line 8: Not sure the wording here is clear. Consider approaching from slightly different angle:
```
Neuropod string type are, similarly to Tensorflow and Pytorch, byte-arrays. A Python3 strings, are unicode strings. To convert a unicode string into a byte-array it needs to be encoded.
```
I think it would be clearer to use numpy encoding methods here instead of an ambiguous `x.astype` (unclear which encoding is used).

BTW, do we still support PY2? If yes, it would mess up a code a little I guess with conditions / six?

source/python/neuropod/backends/neuropod_executor.py
```diff
@@ -175,7 +175,7 @@ def infer(self, inputs):
         # Convert unicode to string
         for k, v in inputs.items():
             if v.dtype.type == np.unicode_:
-                inputs[k] = v.astype("str")
+                inputs[k] = v.astype(np.string_)
```
selitvin, line 178: Should we use an explicit encoding function here and elsewhere?

VivekPanyam:
Superseded by #527

PR #454, opened 2020-10-29
Remove backend loading code that is no longer necessary

### LINKED ISSUE #412: [Java API] Remove isTestMode
#405 added an `isTestMode` flag that causes backend overrides to be specified when loading a model.

This should be refactored to be more generic and, ideally, let us pass in a map of overrides from Java.

https://github.com/uber/neuropod/blob/b3f5eb3e7d7bd01d00bc8445139d395b006db973/source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc#L48-L55

VivekPanyam:
### Summary:
Because of #447 and #451, we no longer need to pass in backend overrides when running tests. This PR removes code that is no longer necessary (including `isTestMode` in the Java bindings).

Closes #412

### Test Plan:

CI


vkuzmin-uber (approved):
Does it fixes https://github.com/uber/neuropod/issues/412 too?

VivekPanyam:
> Does it fixes #412 too?

This PR removes `isTestMode` so it fixes #412, we should probably create another issue to let users pass in backend overrides from Java if they need to.

That feature is much less important now though because of the refactors to how backend loading works (#447, #451)

PR #455, opened 2020-10-29
[TF] Add runtime support for TF 2.3.0

VivekPanyam:
### Summary:

This PR replaces our partial integration with TensorFlow 2.2.0 with a more broad integration of TF 2.3.0.

Since there are now official builds of libtensorflow for TF 2.x, we can use those directly instead of versions I manually built for TF 2.2.0.

_Note: There is still work to be done on the packaging side with TensorFlow 2.x; this PR just allows us to run models with the TF 2.3.0 runtime_

### Test Plan:

CI

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -190,23 +190,23 @@ steps:
     retry:
       automatic: true
 
-  - label: ":docker: CPU Tests (test-cpu-variant-tf-2_2_0-torch-1_5_0-py3_8)"
+  - label: ":docker: CPU Tests (test-cpu-variant-tf-2_3_0-torch-1_5_0-py3_8)"
     timeout_in_minutes: 60
     agents:
       queue: private-default
     command: build/ci/buildkite_build.sh
     plugins:
       - docker-compose#v3.1.0:
-          build: test-cpu-variant-tf-2_2_0-torch-1_5_0-py3_8
+          build: test-cpu-variant-tf-2_3_0-torch-1_5_0-py3_8
           config: docker-compose.test.yml
           image-repository: 027047743804.dkr.ecr.us-east-2.amazonaws.com/uber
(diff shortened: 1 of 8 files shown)


VivekPanyam:
Waiting for TF 2.4.0 for two reasons:

**The CPU tests** fail because libtensorflow for 2.x.x is only built in a manylinux2010 compatible way for CUDA builds:
```sh
  BAZEL_OPTS="--config=opt --cxxopt=-D_GLIBCXX_USE_CXX11_ABI=0"
  export CC_OPT_FLAGS="-mavx -msse4.2"
  if [ "${TF_NEED_CUDA}" == "1" ]; then
    BAZEL_OPTS="${BAZEL_OPTS} --config=cuda --crosstool_top=//third_party/toolchains/preconfig/ubuntu16.04/gcc7_manylinux2010-nvcc-cuda10.1:toolchain"
    export TF_NEED_ROCM=0
  fi
```
See [here](https://github.com/tensorflow/tensorflow/blob/fcc4b966f1265f466e82617020af93670141b009/tensorflow/tools/ci_build/builds/libtensorflow.sh#L54-L59) for the relevant code in TF.

In order for us to fix that issue for TF 2.3.0, we'd need to use our own builds of TF instead of the official release (which is not ideal). Will create an issue and/or PR in the TF repo about this and mention it here.

**The GPU tests** with TF 2.3.0 fails on K80s in CI with

```
2020-10-30 00:31:56.945038: E tensorflow/core/common_runtime/session.cc:91] Failed to create session: Internal: CUDA runtime implicit initialization on GPU:0 failed. Status: device kernel image is invalid
```

The following issue seems related: https://github.com/tensorflow/tensorflow/issues/41990#issuecomment-686336152

Based on https://github.com/tensorflow/tensorflow/commit/32f901f618dfd8dbafdae2187553ff6bcd44de80 and [this annoucement](https://groups.google.com/a/tensorflow.org/g/developers/c/uCKQeiTA3u8/m/xDBWKcrUAgAJ), this issue should be fixed in TF 2.4.0

VivekPanyam:
Issue for manylinux2010 compatibility: https://github.com/tensorflow/tensorflow/issues/45092

VivekPanyam:
Superseded by https://github.com/uber/neuropod/pull/492

PR #459, opened 2020-11-23
[TF] Add runtime support for TF 2.2.0 on Mac

VivekPanyam:
### Summary:
This PR enables us to run models with TF 2.2.0 on Mac (we already had TF 2.2.0 runtime support on Linux).

_Note: There are no official builds of libtensorflow for TF 2.2.0 so we're using binaries I built from source. These will be replaced with official builds from TF 2.3.0 or TF 2.4.0 once they're available. See #455 for more details._ 

### Test Plan:

CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,18 +18,23 @@ matrix:
     # A single linux build to make sure that the build scripts work correctly outside of docker
     - os: linux
     - os: osx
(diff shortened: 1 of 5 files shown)


sepo-eng (approved):
LGTM

Issue #434, opened 2020-09-16
[Java API] Validate that there is a check for an exception after all JNI calls that can raise an exception


vkuzmin-uber:
I looked at "Best practices for using the Java Native Interface" https://developer.ibm.com/technologies/java/articles/j-jni/ and found reasonable advises:

1. "For new JNI code validate that there is a check for an exception after all JNI calls that can raise an exception."
Current Neuropod JNI code doesn't have any "env‑>ExceptionCheck". Except a check, JNI code should clear it after all. Need to use some consistent approach for every JNI call that may throw exception.

Note, I looked at Tensorflow Java/JNI code and found that its JNI exception handling is not consistent, it only checks  for env‑>ExceptionCheck and never clears it.

2. There are JVM flags dedicated for tracing and checking JNI code. Consider what JNI check options available in target JVM (‑Xcheck:jni).  Note source/neuropod/bindings/java/java_build_defs.bzl already uses some settings for "bazel errorprone plugin" (for instance 
 "-Xep:ExpectedExceptionChecker:ERROR"). Need to clarify if it works and if it can report error/warn on current absence of JNI ExceptionCheck calls.

Note that current absence of ExceptionCheck is not crucial because in most of the cases exception means that JVM is already in a bad state. But we are going to use it in a critical service with a high load, at scale, with many dependencies. It is critical that Neuropod Java API is following modern best practices.

PR #462, opened 2020-11-30
[Java API] Add ExceptionCheck when exception is expected after JNI call

vkuzmin-uber:
### Summary:
Most of JNI calls may cause Java Exception that is not thrown automatically in JNI code. We already check results of calls in some cases. This change also adds ExceptionCheck where this is reasonable. If Java exception detected, we print it, clear and throw custom exception. Issue #434 

### Test Plan:
CI


vkuzmin-uber:
@VivekPanyam Vivek, I updated one file only to sync with you if this is a way to deal with it:
- Use neuropod::jni explicitly when it is better for readability
- using neuropod::jni::<entity> otherwise

If this is good, I will update rest then (and neuropod::jni::toUpdate to neuropod::jni::to_update, not sure why linter forgives it)

VivekPanyam (approved):
LGTM!

(We should also probably work on improving test coverage for the Java bindings at some point. It's okay for now as the Java bindings are experimental, but we should improve coverage as we get closer to marking them as prod ready)

PR #463, opened 2020-11-30
Fix some of NOLINTNEXTLINE(google-build-using-namespace)

vkuzmin-uber:
### Summary:
After we enabled Clang-Tidy it reports warning on:

using namespace neuropod::jni;

because including all is a "mauvais ton". This change fixes it in one of JNI file.

### Test Plan:

CI

### DIFF
--- a/source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
+++ b/source/neuropod/bindings/java/src/main/native/com_uber_neuropod_Neuropod.cc
@@ -30,13 +30,14 @@ limitations under the License.
 
 #include <jni.h>
 
+namespace njni = neuropod::jni;
 
 namespace
 {
 
 jobject toJavaTensorSpecList(JNIEnv *env, const std::vector<neuropod::TensorSpec> &specs)
 {
-    jobject ret = env->NewObject(java_util_ArrayList, java_util_ArrayList_, specs.size());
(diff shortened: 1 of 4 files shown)


vkuzmin-uber:
@VivekPanyam Vivek, I updated one file only to sync with you if this is a way to deal with it:
- Use neuropod::jni explicitly when it is better for readability
- using neuropod::jni::"entity" otherwise

If this is good, I will update rest then (and neuropod::jni::toUpdate to neuropod::jni::to_update, not sure why linter forgives it)

VivekPanyam:
@vkuzmin-uber that solution makes sense to me.

You could also use an alias if you want to shorten `neuropod::jni`. Maybe something like
```
namespace njni = neuropod::jni;
```

(For anyone reading, here's some more guidance on namespaces: https://google.github.io/styleguide/cppguide.html#Namespaces)

vkuzmin-uber:
> namespace njni = neuropod::jni;

It makes sense. I will use this strategy, shortcut and use it as prefix for every imported name.

VivekPanyam (approved):
LGTM!

> I updated one file only to sync with you if this is a way to deal with it
> [...]
> If this is good, I will update rest then

@vkuzmin-uber did you made all the changes you wanted to?

(I just wanted to confirm before I merge this PR)

vkuzmin-uber:
> (I just wanted to confirm before I merge this PR)

Yes, this is it for now

PR #471, opened 2021-01-14
[Python] Add initial support for asyncio

VivekPanyam:
### Summary:

Add initial support for running infer using `asyncio`.

This implementation currently blocks the event loop so it shouldn't be used just yet

### Test Plan:

Added a test + CI

### DIFF
--- a/source/python/neuropod/tests/test_async_infer.py
+++ b/source/python/neuropod/tests/test_async_infer.py
@@ -1,0 +1,94 @@
+# Copyright (c) 2021 UATC, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


source/python/neuropod/tests/test_async_infer.py
```diff
@@ -0,0 +1,94 @@
...
+
+            # Sample input data
+            input_data = {
+                "x": np.array([0.5], dtype=np.float32),
+                "y": np.array([1.5], dtype=np.float32),
+            }
+
+            with load_neuropod(path) as model:
+                # Async infer
+                result = loop.run_until_complete(model.infer_async(input_data))
```
vkuzmin-uber, line 87: close loop? Or use "run" that cares about it (I see it has "debug" that maybe good for test) https://docs.python.org/3/library/asyncio-task.html#running-an-asyncio-program

vkuzmin-uber (approved):
LGTM. Left comment, I guess there is a resource leak w/o explicit close of loop

PR #475, opened 2021-01-26
Fix pip bootstrapping on Python 2.7

VivekPanyam:
### Summary:

The latest version of pip drops support for Python 2.7 (https://github.com/pypa/pip/issues/6148).

This PR modifies `install_python_deps.sh` to install a Python 2.7 compatible version of pip.

We should also explore dropping Python 2.7 support in the future as it's been over a year since its EOL.

### Test Plan:

CI

### DIFF
--- a/build/install_python_deps.sh
+++ b/build/install_python_deps.sh
@@ -5,7 +5,7 @@ set -e
 NEUROPOD_PYTHON_BINARY="python${NEUROPOD_PYTHON_VERSION}"
 
 # Install pip
-wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+wget https://bootstrap.pypa.io/2.7/get-pip.py -O /tmp/get-pip.py
 ${NEUROPOD_PYTHON_BINARY} /tmp/get-pip.py
 
 # Setup a virtualenv


build/install_python_deps.sh
```diff
@@ -5,7 +5,7 @@ set -e
 NEUROPOD_PYTHON_BINARY="python${NEUROPOD_PYTHON_VERSION}"
 
 # Install pip
-wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+wget https://bootstrap.pypa.io/2.7/get-pip.py -O /tmp/get-pip.py
```
vkuzmin-uber, line 8: 1. I see it is used at build/ci/install_lint_deps.sh too. 
2. Does it make sense to cache it? like:
```
tools/lint.sh-    if [ ! -f "/tmp/buildifier" ]; then
tools/lint.sh:        wget https://github.com/bazelbuild/buildtools/releases/download/2.2.1/buildifier -O /tmp/buildifier
tools/lint.sh-        chmod +x /tmp/buildifier
tools/lint.sh-    fi
```

or it is not used often?

vkuzmin-uber (approved):
Left comment, though you can go ahead if you need to update here only

build/install_python_deps.sh
VivekPanyam: Not used often. The reason I built the caching in the lint script above (and some other places) is because it's used a lot

PR #476, opened 2021-01-26
Allow users to catch OPE disconnect errors

### LINKED ISSUE #395: Recovery from: "libc++abi.dylib: terminating with uncaught exception of type std::runtime_error"
## Bug
I am testing inference with 4 OPE instances. I am using C++ library from CGo service.

Under high load I observed main process termination:

"libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Neuropod Error: Timed out waiting for a response from worker process. Didn't receive a message in 5000ms, but expected a heartbeat every 2000ms."

The caller has try/catch arounf inference:

```
    try {
       auto output = neuropod->infer(valueMap);
    } catch (const std::exception& e) {
 ...
   }
```

and I know that it catches exception in case of IPE (received TF errors). But it seems this is not the case for OPE. 

Regardless of what caused it, there are 2 questions:

1. How can I catch this exception?
2. If by some reason worker process died, or something else, can I release and load neuropod again to recover? 

## To Reproduce

Load system with high load or somehow else make that parent process doesn't send HeartBeat in time.
UPDATE:
Worker process died. Kill Worker to reproduce.

## Expected behavior

Exception should be caught by caller process.

## Environment

 - Neuropod Version (e.g., 0.2.0): 0.2.0
 - OS (Linux, macOS):
 - Language (Python, C++, Go bindings): CGo, with C bridge.
 - Python version: 
 - Using OPE: yes
 
## Additional context

I suspect that the exception in OPE is not propagated between threads: 

C++ supports
```
std::current_exception();
std::rethrow_exception(teptr);
```
I grep-ed neuropod and don't see these calls.

UPDATE:
If parent process detect timeout, it reports error and exits. It considers this as critical, non-recoverable problem. Though, in production Worker can dies because of OOM or bugs/leaks in platform. Parent process should be able to stay running and handle it.

VivekPanyam:
### Summary:

Fixes #395.

Previously, OPE disconnects were not catchable by user code (as an error was thrown in another thread).

This PR refactors some of the IPC message queue send and recv code to handle disconnects in a way that gives the user more control

### Test Plan:

CI

### DIFF
--- a/build/install_python_deps.sh
+++ b/build/install_python_deps.sh
@@ -5,7 +5,7 @@ set -e
 NEUROPOD_PYTHON_BINARY="python${NEUROPOD_PYTHON_VERSION}"
 
 # Install pip
-wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+wget https://bootstrap.pypa.io/2.7/get-pip.py -O /tmp/get-pip.py
 ${NEUROPOD_PYTHON_BINARY} /tmp/get-pip.py
 
 # Setup a virtualenv
--- a/source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
+++ b/source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
@@ -76,10 +76,20 @@ void IPCMessageQueue<UserPayloadType>::read_worker_loop()
         if (!successful_read)
         {
             // We timed out
-            NEUROPOD_ERROR("Timed out waiting for a response from worker process. "
-                           "Didn't receive a message in {}ms, but expected a heartbeat every {}ms.",
-                           detail::MESSAGE_TIMEOUT_MS,
-                           detail::HEARTBEAT_INTERVAL_MS);
+            SPDLOG_ERROR("Timed out waiting for a response from worker process. "
+                         "Didn't receive a message in {}ms, but expected a heartbeat every {}ms.",
+                         detail::MESSAGE_TIMEOUT_MS,
+                         detail::HEARTBEAT_INTERVAL_MS);
+
+            // Set a flag so that any pending writers know
+            lost_heartbeat_.store(true, std::memory_order_relaxed);
+
+            // Let any pending readers know
+            // (Because recv isn't threadsafe, there should only be one)
+            out_queue_.try_emplace(nullptr);
+
+            // Shutdown this thread
+            break;
         }
 
         if (received->type == detail::USER_PAYLOAD)
@@ -152,7 +162,32 @@ void IPCMessageQueue<UserPayloadType>::send_message(const WireFormat &msg)
         SPDLOG_TRACE("OPE: Sending IPC control message {}.", msg.type);
     }
 
-    send_queue_->send(&msg, sizeof(msg), 0);
+    // Send w/ timeout + heartbeat check
+    while (true)
+    {
+        auto timeout_at = boost::interprocess::microsec_clock::universal_time() +
+                          boost::posix_time::milliseconds(detail::HEARTBEAT_INTERVAL_MS);
+
+        if (send_queue_->timed_send(&msg, sizeof(msg), 0, timeout_at))
+        {
+            // Successfully sent
+            break;
+        }
+
+        // Make sure the worker process is still alive
+        throw_if_lost_heartbeat();
+    }
+}
+
+// Throw an error if we lost communication with the other process
+template <typename UserPayloadType>
+void IPCMessageQueue<UserPayloadType>::throw_if_lost_heartbeat()
+{
+    if (lost_heartbeat_.load(std::memory_order_relaxed))
+    {
+        // We lost the heartbeat so let's throw an error here (where the user can catch it)
+        NEUROPOD_ERROR("OPE lost communication with the other process. See logs for more details.");
+    }
 }
 
 template <typename UserPayloadType>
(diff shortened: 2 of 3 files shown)


source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
```diff
@@ -152,7 +162,32 @@ void IPCMessageQueue<UserPayloadType>::send_message(const WireFormat &msg)
...
+}
+
+// Throw an error if we lost communication with the other process
+template <typename UserPayloadType>
+void IPCMessageQueue<UserPayloadType>::throw_if_lost_heartbeat()
+{
+    if (lost_heartbeat_.load(std::memory_order_relaxed))
+    {
+        // We lost the heartbeat so let's throw an error here (where the user can catch it)
+        NEUROPOD_ERROR("OPE lost communication with the other process. See logs for more details.");
```
vkuzmin-uber, line 189: This can be considered separately, just mention here. This will throw exception:

```
 template <typename... Params>
 [[noreturn]] void throw_error(const char *file, int line, const char *function, Params &&... params)
 {
     // Log the error
     // (this is effectively what the SPDLOG_ERROR macro expands to)
     spdlog::default_logger_raw()->log(
         spdlog::source_loc{file, line, function}, spdlog::level::err, std::forward<Params>(params)...);

     // Throw a runtime error
     throw std::runtime_error("Neuropod Error: " + fmt::format(std::forward<Params>(params)...));
 }
 ```
 
 If parent/core want to distinguish exceptions, this will need to parse/compare exception message. It would be useful to have some way to categorize it, I don't think that attribute "category" is necessary but maybe having some "tags" as a contract may help.

File: build/install_python_deps.sh [hunk 2]
```diff
@@ -5,7 +5,7 @@ set -e
 NEUROPOD_PYTHON_BINARY="python${NEUROPOD_PYTHON_VERSION}"
 
 # Install pip
-wget https://bootstrap.pypa.io/get-pip.py -O /tmp/get-pip.py
+wget https://bootstrap.pypa.io/2.7/get-pip.py -O /tmp/get-pip.py
```
vkuzmin-uber, line 8: just note: It was merged, need to rebase.

source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
```diff
@@ -163,6 +198,7 @@ IPCMessageQueue<UserPayloadType>::IPCMessageQueue(const std::string &control_que
       recv_queue_(detail::make_recv_queue<UserPayloadType>(control_queue_name_, type)),
       heartbeat_controller_(stdx::make_unique<detail::HeartbeatController>(*this)),
       transferrable_controller_(stdx::make_unique<detail::TransferrableController>()),
+      lost_heartbeat_(false),
```
vkuzmin-uber, line 201: I don't see a full picture, maybe you checked this already.

This will be set to true on failure and never reset again. I guess the idea is that the given Neuropod object is not re-usable (that is fine). 

What happens with next "infer", isn't it causing other crash? Next infer may happen because of race, when it happend before Parent received exception and can stop serving.

vkuzmin-uber (commented):
Left comment - concern on what is the state of Neuropod after that and how next infer is handled. Maybe Unit Test may cover it and show that it doesn't crash parent again.

source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
```diff
@@ -152,7 +162,32 @@ void IPCMessageQueue<UserPayloadType>::send_message(const WireFormat &msg)
         SPDLOG_TRACE("OPE: Sending IPC control message {}.", msg.type);
     }
 
-    send_queue_->send(&msg, sizeof(msg), 0);
+    // Send w/ timeout + heartbeat check
+    while (true)
```
mincomp, line 166: can you explain a bit why there's a while loop? it seems if send does not time out and heartbeat is not lost we'll indefinitely retry?

source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
VivekPanyam: > I guess the idea is that the given Neuropod object is not re-usable (that is fine).

Yes, this was the case before and continues to be the case. Previously, a disconnect would cause an error to be thrown in another thread (and take down the whole application). Now, a disconnect will throw an error in the current thread, but the model is no longer usable. 

> Next infer may happen because of race, when it happend before Parent received exception and can stop serving.

As we've talked about a few times before, `infer` is _not_ threadsafe. `infer` should only be called by one thread at a time. If you're using a single model across threads with locking, I'd recommend unloading the model when you get a disconnect exception.

If you do call `infer` again, it will throw an error again (after potentially blocking for a bit).

source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
VivekPanyam: Makes sense - I can make the exception distinguishable

source/neuropod/multiprocess/mq/ipc_message_queue_impl.hh
VivekPanyam: >  if send does not time out and heartbeat is not lost we'll indefinitely retry?

If send does not time out, we'll return (lines 173 and 174).

If send times out, we'll keep retrying until it works or we lose the heartbeat. This is the same behavior as before (except the whole application doesn't terminate when we lose a heartbeat).

VivekPanyam (commented):
> Left comment - concern on what is the state of Neuropod after that and how next infer is handled. Maybe Unit Test may cover it and show that it doesn't crash parent again.

The model is no longer valid after OPE disconnects and `infer` throws an error. As such, there should not be another call to `infer`. (If there is another call to `infer`, it will throw another error.)

The purpose of this PR is to prevent an OPE disconnect from taking down the application - not to make models reusable once the worker process disconnects.

vkuzmin-uber (approved):

Issue #486, opened 2021-04-06
Add support for tf.SavedModel format


tgaddair:
In TensorFlow v2, the recommended storage format for trained models is the [SavedModel](https://www.tensorflow.org/guide/saved_model) format, but this is not currently supported by Neuropod. Being able to support this with an API similar to Keras, but from a `.pb` file saved on disk, would be great.

VivekPanyam:
I agree we should add support for TF SavedModels. If anyone's willing to help with implementing this, please comment here.

Also, to clarify, Neuropod does support packaging TF `.pb` files. `create_tensorflow_neuropod` lets you pass in a [GraphDef](https://neuropod.ai/docs/stable/packagers/tensorflow/#graph_def) or a [frozen graph](https://neuropod.ai/docs/stable/packagers/tensorflow/#frozen_graph_path) (a `.pb` file you've already exported)

VivekPanyam:
I spent a bunch of time on this over the last 9 hours and now have an initial SavedModel implementation that passes tests from both Python and C++ (with TF 2.2)!

It was fairly complex to get working, but I think there's a reasonable path forward.

I'll post more context in this thread tomorrow.

vkuzmin-uber:
Vivek, 

Do you use it to convert to frozen graph or use it to initialize?

FYI: I did conversion once and when looked for generic solution found this code  https://github.com/ogemarques/models/blob/master/research/tensorrt/tensorrt.py#L163

helinwang:
Hey Vivek, this is great, thank you! Looking forward for it.

VivekPanyam:
Hey everyone,

Here's the context I mentioned above:

## Context

### Why doesn't Neuropod already have support for `SavedModel`s?

The initial TF implementation for Neuropod was built when ATG was using TF (1.x) frozen graphs directly. All of our models required only frozen graphs to run inference so SavedModel support wasn't a pressing need.

I remember spending a decent amount of time exploring SavedModel support in the past, but there was significant overhead to getting it working well in Neuropod (compatibility with multiple TF versions, relatively complex integration, making it easy for users to export models without having to provide the same set of information multiple times, etc).

Our existing solution fit our needs (and the needs of all of our users as far as we were aware) so the implementation and design effort wasn't worth it at the time. ATG also started focusing on PyTorch for newer models so we didn't have a compelling reason to make major changes to something that wasn't broken.

### Why not use the built-in `LoadSavedModel` in Tensorflow?

[...]

PR #487, opened 2021-04-20
[TF] Add initial support for SavedModel

VivekPanyam:
### Summary:

This PR adds initial support for packaging and loading TensorFlow `SavedModel`s

Doing so is not entirely straightforward as described in my comment [here](https://github.com/uber/neuropod/issues/486#issuecomment-820228901).

As mentioned in that comment, we can't use TF's `LoadSavedModel` directly because of device issues with multiple TF models running in one process. As a solution, this PR pulls in the [TF 2.4 SavedModel loading code](https://github.com/tensorflow/tensorflow/tree/r2.4/tensorflow/cc/saved_model) and modifies it to handle devices correctly.

The comment linked above goes into more detail on what the modifications are and why they're needed. There is also a README file in this PR that includes a few more details. 

### Test Plan:

CI + added tests

Note: this is currently only tested with TF 2.x

### DIFF
--- a/source/neuropod/backends/tensorflow/saved_model/loader.cc
+++ b/source/neuropod/backends/tensorflow/saved_model/loader.cc
@@ -1,0 +1,329 @@
+/* Copyright 2016 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// Mostly from https://github.com/tensorflow/tensorflow/blob/r2.4/tensorflow/cc/saved_model/loader.cc
+// with some modifications
+
+#include "neuropod/backends/tensorflow/saved_model/loader.h"
+
+#include "neuropod/backends/tensorflow/saved_model/constants.h"
+#include "neuropod/backends/tensorflow/saved_model/loader_util.h"
+#include "neuropod/backends/tensorflow/saved_model/reader.h"
+#include "neuropod/backends/tensorflow/tf_utils.hh"
+#include "tensorflow/core/framework/attr_value.pb.h"
+#include "tensorflow/core/framework/function.pb.h"
+#include "tensorflow/core/framework/node_def.pb.h"
+#include "tensorflow/core/framework/tensor.pb.h"
+#include "tensorflow/core/lib/core/errors.h"
+#include "tensorflow/core/lib/io/path.h"
+#include "tensorflow/core/lib/monitoring/counter.h"
+#include "tensorflow/core/lib/monitoring/sampler.h"
+#include "tensorflow/core/lib/strings/str_util.h"
+#include "tensorflow/core/lib/strings/strcat.h"
+#include "tensorflow/core/lib/strings/stringprintf.h"
+#include "tensorflow/core/platform/env.h"
+#include "tensorflow/core/protobuf/meta_graph.pb.h"
+#include "tensorflow/core/protobuf/saver.pb.h"
+#include "tensorflow/core/public/session.h"
+#include "tensorflow/core/public/session_options.h"
+#include "tensorflow/core/public/version.h"
+
+#include <unordered_set>
+
+namespace tensorflow
+{
+namespace
+{
+
+// Ensure that constant tensors loaded from the saved model have valid shape.
+// Also ensure that constant nodes have a value assigned to them.
+// TODO(b/154763635): this is temporary and will be replaced with a better audit
+static Status ValidateNode(const NodeDef &node)
+{
+    const auto node_iterator = node.attr().find("value");
+    if (node_iterator != node.attr().end())
+    {
+        AttrValue node_value = node_iterator->second;
+        if (node_value.has_tensor())
+        {
+            const PartialTensorShape node_shape(node_value.tensor().tensor_shape());
+            if (node_shape.num_elements() < 0)
+            {
+                return errors::FailedPrecondition("Saved model contains node \"",
+                                                  node.name(),
+                                                  "\" (op \"",
+                                                  node.op(),
+                                                  "\") which initializes from a tensor with ",
+                                                  node_shape.num_elements(),
+                                                  " elements");
+            }
+        }
+    }
+    else if (node.op() == "Const")
+    {
+        return errors::FailedPrecondition("Saved model contains node \"",
+                                          node.name(),
+                                          "\" which is a constant tensor but no value has been provided");
+    }
+    return Status::OK();
+}
+
+static Status ValidateSavedTensors(const GraphDef &graph_def)
+{
+    for (const auto &node : graph_def.node())
+    {
+        TF_RETURN_IF_ERROR(ValidateNode(node));
+    }
+
+    if (graph_def.has_library())
+    {
+        const FunctionDefLibrary &library = graph_def.library();
+        for (const auto &function : library.function())
+        {
+            for (const auto &node : function.node_def())
+            {
+                TF_RETURN_IF_ERROR(ValidateNode(node));
+            }
+        }
+    }
+
+    return Status::OK();
+}
+
+// NEUROPOD MODIFICATION (to be compatible with TF 1.x and TF 2.x)
+#if TF_MAJOR_VERSION > 1
+using TFStringType = tensorflow::tstring;
+#else
+using TFStringType = std::string;
+#endif
+
+Tensor CreateStringTensor(const string &value)
+{
+    Tensor tensor(DT_STRING, TensorShape({}));
+    tensor.scalar<TFStringType>()() = value;
+    return tensor;
+}
+
+void AddAssetsTensorsToInputs(const StringPiece                       export_dir,
+                              const std::vector<AssetFileDef> &       asset_file_defs,
+                              std::vector<std::pair<string, Tensor>> *inputs)
+{
+    if (asset_file_defs.empty())
+    {
+        return;
+    }
+    for (auto &asset_file_def : asset_file_defs)
+    {
+        Tensor assets_file_path_tensor =
+            CreateStringTensor(io::JoinPath(export_dir, kSavedModelAssetsDirectory, asset_file_def.filename()));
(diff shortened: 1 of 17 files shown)


VivekPanyam:
cc @csashiuber @thuningxu @tgaddair @helinwang

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
```diff
@@ -0,0 +1,115 @@
...
+                tf.TensorSpec(shape=None, dtype=tf.float32),
+            ]
+        )
+        def add(self, x, y):
+            return {"out": x + y}
+
+    return Adder()
+
+
+@unittest.skipIf(tf.__version__[0] != "2", "Skipping TF 2.x tests for TF 1.x")
```
sepo-eng, line 43: Why skip tests for TF 1.x? In the TF saved_model code, you made changes to specifically support multiple versions.

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
```diff
@@ -0,0 +1,115 @@
...
+class TestTensorflowV2Packaging(unittest.TestCase):
```
sepo-eng, line 44: Can you add tests for the 1 of logic for graph, saved_model and trackable_obj?

source/neuropod/backends/tensorflow/tf_backend.cc
```diff
@@ -152,47 +145,40 @@ TensorflowNeuropodBackend::TensorflowNeuropodBackend(const std::string &neuropod
...
-                else
-                {
-                    NEUROPOD_ERROR("Failed to load custom op. Error from dlopen: {}", err);
-                }
-            }
+    session_ = std::move(bundle.session);
 
-            loaded_op_hashes.insert(hash);
-        }
+    // Get signatures
```
sepo-eng, line 157: It's not clear what these signatures represent, can you expand the comment to include an explanation?

source/neuropod/backends/tensorflow/saved_model/loader.cc
```diff
@@ -0,0 +1,329 @@
...
+    Session *session_p = nullptr;
+    TF_RETURN_IF_ERROR(NewSession(session_options, &session_p));
+    session->reset(session_p);
+
+    GraphDef &graph = *meta_graph.mutable_graph_def();
+    TF_RETURN_IF_ERROR(ValidateSavedTensors(graph));
+
+    // NEUROPOD MODIFICATION
+    // This moves the graph to the appropriate device before loading it
+    neuropod::move_graph_to_device(graph, *(session->get()), 0);
```
sepo-eng, line 283: We're creating a confusing dependency model here, by introducing a neuropod namespace dependency on tensorflow. It might be preferable to move the move_graph_to_device logic in the tensorflow namespace, and have the graph loader depend on this logic instead, so the dependency chain remains linear.

source/neuropod/backends/tensorflow/tf_backend.cc
```diff
@@ -250,15 +202,57 @@ void TensorflowNeuropodBackend::load_model_internal()
...
+{
+    // Load custom ops (if any)
+    for (const auto &item : model_config_->custom_ops)
+    {
+        const auto path = "0/ops/" + item;
+        const auto hash = loader_->get_hash_for_file(path);
+
+        // Don't load a custom op if we've already loaded it
+        std::lock_guard<std::mutex> lock(loaded_op_mutex);
+        if (loaded_op_hashes.count(hash) == 0)
```
sepo-eng, line 222: Can you comment on how this loaded_op_hashes is used?

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
```diff
@@ -0,0 +1,115 @@
...
+        with TemporaryDirectory() as test_dir:
+            neuropod_path = os.path.join(test_dir, "test_neuropod")
+
+            # Get the input/output spec along with test data
+            kwargs.update(get_addition_model_spec(do_fail=do_fail))
+
+            # `create_tensorflow_neuropod` runs inference with the test data immediately
+            # after creating the neuropod. Raises a ValueError if the model output
+            # does not match the expected output.
+            create_tensorflow_neuropod(
```
sepo-eng, line 55: Can you add tests for use of saved_model? It seems like the logic added in this change isn't tested. check_addition_model will test the forward call right?

sepo-eng (commented):
The changes themselves LGTM, except for missing native tests and some potential issues I've outlined in the comments. 

I'm mainly concerned about maintaining what is essentially a copy of Tensorflow code. Is it possible to support these changes in TF instead?

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
VivekPanyam: AFAIK, the way to create a TF saved model from Python is different in TF 1.x vs 2.x.

The pressing usecase is for TF 2.x so maybe we can just create an issue to add savdmodel tests for TF 1.x?

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
VivekPanyam: `graph_def` is tested [here](https://github.com/uber/neuropod/blob/01a87fea514249a268b4ad996d477f64286a9a75/source/python/neuropod/backends/tensorflow/test/test_tensorflow_packaging.py#L67) for TF 1.x. Graph defs aren't really natural to deal with in TF 2.x (and aren't really recommended afaik) so there isn't a test for it here.

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
VivekPanyam: > Can you add tests for use of saved_model? It seems like the logic added in this change isn't tested.

Passing in `trackable_obj` exercises the same code as passing in `saved_model_dir` at inference time.

The only line in the packaging code that isn't exercised (because we're not testing `saved_model_dir`) is 

https://github.com/uber/neuropod/blob/38091314e8544abf0b50403344a9b1b1f3fa4a6f/source/python/neuropod/backends/tensorflow/packager.py#L96-L97

(you can confirm by looking at code coverage if you'd like)

I'll add a test for it though :)

> check_addition_model will test the forward call right?

Yes. `create_tensorflow_neuropod` also runs inference (see the comment on the lines above this one):

```py
            # `create_tensorflow_neuropod` runs inference with the test data immediately
            # after creating the neuropod. Raises a ValueError if the model output
            # does not match the expected output.
```

source/neuropod/backends/tensorflow/saved_model/loader.cc
VivekPanyam: I'd like to avoid adding things to the `tensorflow` namespace that don't absolutely need to be there. There are some situations where it's necessary (like friend classes that need to be in particular namespaces), but I don't think we need to do that here. Ideally the `tensorflow` namespace would only contain things directly from the TF codebase.

I think this also helps make it clear what's Neuropod code and what came from TF.

I haven't seen anything about enforcing linear dependency chains in C++ namespaces, but it's possible I'm missing something. Do you have a style guide reference for this?

The [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Namespaces) doesn't mention anything about this as far as I can tell.

source/neuropod/backends/tensorflow/tf_backend.cc
VivekPanyam: This is preexisting code that I just moved around (unrelated to savedmodel). Is it okay if I clarify in a follow up PR if necessary?

The functionality seems pretty clear to me based on the comments and the code in lines 214 to 239, but I've spent a lot of time in this codebase so I could think it's more clear than it actually is haha.

I'll follow up offline

source/neuropod/backends/tensorflow/tf_backend.cc
VivekPanyam: Will do. I'll also add a reference to the TF docs

VivekPanyam (commented):
Thanks for the review!

> The changes themselves LGTM, except for missing native tests and some potential issues I've outlined in the comments.

Responded to the comments. The python tests exercise the native inference code as well as the python inference code. (CI runs all python tests twice - once with native bindings, once with the python implementation). This lets us avoid writing two copies of effectively the same test.

> I'm mainly concerned about maintaining what is essentially a copy of Tensorflow code. 

We spent a while looking into alternate solutions (when we tried to do this in the past and again now), but there aren't any great ones that we found. The one in this PR seems to be the cleanest and most maintainable. Please see the PR description and [this](https://github.com/uber/neuropod/issues/486#issuecomment-820228901) comment for more details.

The code from TF should not need to be changed after bringing it in initially (and if we do need to upgrade it, all changes from the original code are quite minor and called out in comments so we should be able to replicate easily). Happy to chat more offline.

> Is it possible to support these changes in TF instead?

Unfortunately it would be quite difficult to move these changes into TF. Lots of reasons, but here are a few:
- Wouldn't work for already released versions of TF (which Neuropod needs to support)
- The way TF handles devices is quite complex and needs to support many different usecases. This approach isn't general enough to pull into TF. See the comment I linked above for more detail
- Even if we got to a solution that was acceptable to pull into TF, this would be a major change and would take a lot of resources to get done

Again, happy to chat more offline if you want

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
sepo-eng: That sounds good to me

source/neuropod/backends/tensorflow/saved_model/loader.cc
sepo-eng: I understand why you would like to limit modifications to tensorflow, but the circular dependency we're introducing, while not breaking is also not ideal. 

I don't have deep knowledge of either codebase so this suggestion might be naive, but this is essentially what I'm getting at:

> I'm mainly concerned about maintaining what is essentially a copy of Tensorflow code. Is it possible to support these changes in TF instead?

Maybe we can avoid this circular dependency as well as the copy of TF code if the move_graph_to_device API is implemented in TF.

sepo-eng: Reading these comments in linear order, and just saw your response, and it seems like something you've already considered. Resolving this conversation.

sepo-eng (approved):

source/neuropod/backends/tensorflow/tf_backend.cc
VivekPanyam: Correct!

source/python/neuropod/tests/test_tensorflow_v2_packaging.py
VivekPanyam: Created https://github.com/uber/neuropod/issues/496

PR #515, opened 2021-11-04
Support StringTensor as model output in Neuropod JNI

voe09:
### Summary:
In neuropod JNI, it does not support STRING_TENSOR as model output. The reason is when it tries to create a Java Tensor from model output, it would check the tensor type. Only numerical tensors are supported and these tensors would be converted to ByteBuffer. However, Java NeuropodTensor itself do support STRING_TENSOR. It can get strings from the tensor data itself. Thus we refactor nativeGetBuffer method. This method would assign a null buffer to bytebuffer if it is a string tensor

### Test Plan:
Unit test

### DIFF
--- a/source/neuropod/bindings/java/BUILD
+++ b/source/neuropod/bindings/java/BUILD
@@ -148,6 +148,24 @@ java_test(
 )
 
 java_test(
+    name = "TFStringsModelTest",
+    srcs = [
+        "src/test/java/com/uber/neuropod/NeuropodStringsModelTest.java",
+        "src/test/java/com/uber/neuropod/TFStringsModelTest.java",
+    ],
+    data = [
+        "//neuropod/tests/test_data",
+    ],
+    javacopts = JAVACOPTS,
+    tags = ["requires_framework_tensorflow"],
+    test_class = "com.uber.neuropod.TFStringsModelTest",
+    deps = [
+        ":neuropod_java_jar",
+        "@junit",
+    ],
+)
+
+java_test(
     name = "TorchscriptAdditionTest",
     srcs = [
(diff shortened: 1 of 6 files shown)


qiyanz (approved):

source/neuropod/bindings/java/BUILD
```diff
@@ -165,6 +183,24 @@ java_test(
...
+    name = "TorchscriptStringsModelTest",
+    srcs = [
+        "src/test/java/com/uber/neuropod/NeuropodStringsModelTest.java",
+        "src/test/java/com/uber/neuropod/TorchscriptStringsModelTest.java",
+    ],
+    data = [
+        "//neuropod/tests/test_data",
+    ],
+    javacopts = JAVACOPTS,
+    tags = ["requires_framework_tensorflow"],
```
VivekPanyam, line 196: Replace with `requires_framework_torchscript`

VivekPanyam (changes requested):
Thanks for the PR! Looks good overall.

I left one comment that needs to be fixed in order for CI to pass. Once CI passes, I'll go ahead and merge the PR. Thanks again!

source/neuropod/bindings/java/BUILD
voe09: Thanks @VivekPanyam I've updated my pr.

voe09:
Hi @VivekPanyam could you help take a look at the diff. I have rebased the master branch and all CI passed. Thanks!

VivekPanyam:
None of the Linux CI has passed yet, but just unblocked it so it should be running now

VivekPanyam (approved):
LGTM!

PR #519, opened 2021-11-05
Use TF2.6.2 in CI TF2.6 Py3.8 Pipeline

voe09:
### Summary:
Keras 2.7 breaks TF2.6.0. The issue is fixed in TF2.6.2. We would ping our TF 2.6 version to https://github.com/tensorflow/tensorflow/releases/tag/v2.6.2 in CI

### DIFF
--- a/.github/workflows/mac_ci.yml
+++ b/.github/workflows/mac_ci.yml
@@ -78,7 +78,7 @@ jobs:
                       python: 3.8
                       test_frameworks: tensorflow
 
-                    - tf: 2.6.0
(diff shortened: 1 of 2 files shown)


VivekPanyam (commented):
Unfortunately, the change to use 2.6.2 isn't that simple. I'm working on a PR that I should have up shortly

voe09:
Vivek is working on pinging TF version. Abandon the PR for now

VivekPanyam:
See https://github.com/uber/neuropod/pull/520

PR #526, opened 2021-12-15
Remove _always_use_native=False

VivekPanyam:
### Summary:

`_always_use_native=False` was always intended to be a workaround (hence the leading `_`) and has had a deprecation warning since #508.

Removing this functionality allows us to delete a lot of code and have a single inference code path across C, C++, Java, Python, and Go. 

### Test Plan:

CI


VivekPanyam:
cc @csashiuber @qiyanz @vkuzmin-uber

qiyanz (approved):

PR #531, opened 2022-02-16
Add support for Torch 1.10.2

VivekPanyam:
### Summary:

This PR adds support for Torch 1.10.2 and adds a build to the CI matrix to test it

### Test Plan:

CI

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -331,6 +331,47 @@ steps:
     retry:
       automatic: true
 
+  - label: ":docker: CPU Tests (test-cpu-variant-tf-2_2_0-torch-1_10_2-py3_8)"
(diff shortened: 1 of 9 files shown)


VivekPanyam:
> Thanks for putting this PR together, @VivekPanyam. Is this good to go, or are there issues that need to be addressed? (never sure with CI what's flaky and what's not)

Need to land #532 first (pip broke on py3.6 because of EOL + a python ppa changed and broke CI). We're running CI on GPU spot instances and they've been getting terminated a lot over the last few days so it's taking a bit longer than expected for builds to pass.

Hopefully we'll be able to get this out by the end of the week

tgaddair:
Awesome, that sounds great. Thanks for the update.

voe09 (approved):

PR #537, opened 2022-03-08
Modify copyright headers and add an AUTHORS file

VivekPanyam:
### Summary:

This PR modifies the copyright headers across Neuropod to more correctly reflect the actual copyright holders of contributions to the project.

(Note that changing the text of the header doesn't change any ownership; it just more clearly reflects that all contributions are not owned by `UATC, LLC`)

A common way to do this is to introduce an AUTHORS file.

Please see https://opensource.google/documentation/reference/releasing/authors for more information about AUTHORS files

### Test Plan:

N/A

### DIFF
--- a/source/neuropod/bindings/java/src/test/java/com/uber/neuropod/PythonStringsModelTest.java
+++ b/source/neuropod/bindings/java/src/test/java/com/uber/neuropod/PythonStringsModelTest.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020 UATC, LLC
+/* Copyright (c) 2022 The Neuropod Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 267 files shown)


source/neuropod/bindings/java/src/test/java/com/uber/neuropod/PythonStringsModelTest.java
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020 UATC, LLC
+/* Copyright (c) 2022 The Neuropod Authors
```
VivekPanyam, line 1: (note that this date was changed from `2020` to `2022` because the file was actually created in 2022, but I listed the wrong date in the original commit)

thuningxu (approved):

Issue #542, opened 2022-04-05
Support Dict[str, Union[List[str], torch.Tensor]] as torchscript model input type


voe09:
Neuropod has supported torch 1.10.2 in its latest release. In Torch 1.10.2, torch.jit introduced the support for Union type. That means it can support Dict[str, Union[List[str], torch.Tensor]] as its model input type.

Currently in neuropod, torchscript backend can only support Dict[str, List[str]] or Dict[str, torch.Tensor]. Related code is https://github.com/uber/neuropod/blob/master/source/neuropod/backends/torchscript/torch_backend.cc#L354

We would like to request support for Dict[str, Union[List[str], torch.Tensor]]

VivekPanyam:
As we talked about offline, this sounds good to me.

Some additional thoughts that you should feel free to ignore (or we should create a separate issue about if they sound useful):

* It would be great to also include the option of a model taking multiple dicts as inputs instead of just one (so if a model takes in a string dict and a tensor dict, Neuropod will supply both). This would have the benefit of supporting older versions of Torch as well. It should be a fairly straightforward change to implement this.
* Neuropod already supports several output types (`Tensor`, `Dict[str, Tensor]`, named tuples, `Dict[str, List[str]]`, tuple of dicts to support multiple return types, etc). Adding the ability to return `Dict[str, Union[List[str], torch.Tensor]]` would be nice if we are going to support that type as an input

These are definitely not required as part of this task (and should not all be put in the same PR), but some things to consider if you're interested in improving the TorchScript model interface.

---

_Permalink version of the code snippet linked in original issue above:_

https://github.com/uber/neuropod/blob/1097df8fc683141db08f8d66c67cae45d0a83a33/source/neuropod/backends/torchscript/torch_backend.cc#L354-L355

PR #543, opened 2022-04-07
Support Dict[str, Union[List[str], torch.Tensor] as model input type for TorchScript 1.10.2 backend

voe09:
### Summary:
See details in 
https://github.com/uber/neuropod/issues/542

### Test Plan:
Unit test


source/neuropod/tests/test_utils.hh
```diff
@@ -209,3 +209,62 @@ void test_strings_model_ope(const std::string &neuropod_path)
     neuropod::Neuropod neuropod(neuropod_path, opts);
     test_strings_model(neuropod);
 }
+
+void test_torchscript_dict_with_union_value_type_model(neuropod::Neuropod &neuropod)
+{
+#if CAFFE2_NIGHTLY_VERSION >= 20220127
```
VivekPanyam, line 215: 1. This file is meant for generic test utils - not ones that are specific to a particular backend. These additions should be either renamed to be more generic (if we can use them in tests for other backends as well) or moved into the torch backend folder.

2. `CAFFE2_NIGHTLY_VERSION` is defined in `torch_tensor.hh`: https://github.com/uber/neuropod/blob/1097df8fc683141db08f8d66c67cae45d0a83a33/source/neuropod/backends/torchscript/torch_tensor.hh#L32-L80 Since we're not including `torch_tensor.hh` here or in `test_torchscript_backend.cc`, `CAFFE2_NIGHTLY_VERSION` is undefined. This means none of your gated test code is included in the build.

A few solutions:

* Rename the functions to be backend agnostic and move the `#if CAFFE2_...` into torch specific files (which include `torch_tensor.hh`)
* Move these functions into the torchscript backend test directory and include `torch_tensor.hh`

To make problems like this easier to catch, we should also build with `-Wundef`. I had a series of PRs converting the Neuropod build to use a very strict set of warnings flags (see [here](https://github.com/uber/neuropod/blob/1097df8fc683141db08f8d66c67cae45d0a83a33/source/bazel/cc.bzl#L4-L26)), but those settings are only used by part of the build at the moment.

VivekPanyam (changes requested):
Thanks for the PR!

Looks good overall, but as you pointed out, the added code is not executed by tests (based on code coverage). See my inline comment explaining why this is happening.

I'd also recommend writing python tests for this PR. The python tests execute the native code path and also include model packaging. They also tend to be simpler to write and maintain than the C++ inference tests that use models in `test_data`.

You can see examples of TorchScript tests [here](https://github.com/uber/neuropod/blob/master/source/python/neuropod/backends/torchscript/test/test_torchscript_packaging.py) and [here](https://github.com/uber/neuropod/blob/master/source/python/neuropod/backends/torchscript/test/test_torchscript_strings.py).

Those files test models with several input and output formats and it should be fairly straightforward to add a test or two there to make sure union works.

VivekPanyam:
One more thing to add: If it's easier, you can just replace the C++ tests you added with Python tests (instead of fixing the C++ tests and adding Python tests as well)

source/neuropod/backends/torchscript/test/torchscript_test_utils.hh
```diff
@@ -0,0 +1,84 @@
...
+#include "gtest/gtest.h"
+#include "neuropod/neuropod.hh"
+#include "neuropod/backends/torchscript/torch_tensor.hh"
+
+#include <string>
+#include <vector>
+
+#include <stdlib.h>
+
+void test_torchscript_dict_with_union_value_type_model(neuropod::Neuropod &neuropod)
```
VivekPanyam, line 27: You can probably just move this function into an anonymous namespace in the `.cc` file

source/python/neuropod/backends/torchscript/test/test_torchscript_dict_with_union_value_type_model.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2020 The Neuropod Authors
```
VivekPanyam, line 1: 2022

source/neuropod/backends/torchscript/test/torchscript_test_utils.hh
```diff
@@ -0,0 +1,84 @@
...
+
+    const std::vector<int64_t> out_shape = output_data->at("c")->as_tensor()->get_dims();
+
+    EXPECT_EQ(output_vec.size(), 6);
+    EXPECT_TRUE(std::equal(output_vec.begin(), output_vec.end(), target));
+    EXPECT_TRUE(out_shape == expected_output_shape);
+#endif
+}
+
+void test_torchscript_dict_with_union_value_type_model(const std::string &neuropod_path)
```
VivekPanyam, line 59: Move the contents of this function directly into the `TEST` block that references it (since it is likely not going to be used anywhere else)

source/neuropod/backends/torchscript/test/torchscript_test_utils.hh
```diff
@@ -0,0 +1,84 @@
...
+    // Should fail because we haven't loaded the model yet
+    EXPECT_ANY_THROW(test_torchscript_dict_with_union_value_type_model(neuropod));
+
+    // Load the model and try again
+    neuropod.load_model();
+    test_torchscript_dict_with_union_value_type_model(neuropod);
+#endif
+}
+
+void test_torchscript_dict_with_union_value_type_model_ope(const std::string &neuropod_path)
```
VivekPanyam, line 76: Move the contents of this function directly into the `TEST` block that references it (since it is likely not going to be used anywhere else)

source/neuropod/backends/torchscript/test/torchscript_test_utils.hh
```diff
@@ -0,0 +1,84 @@
...
+{
+#if CAFFE2_NIGHTLY_VERSION >= 20220127
+    neuropod::RuntimeOptions opts;
+    opts.use_ope = true;
+    neuropod::Neuropod neuropod(neuropod_path, opts);
+    test_torchscript_dict_with_union_value_type_model(neuropod);
+#endif
+}
```
VivekPanyam, line 84: After the above changes, you should be able to delete this file

VivekPanyam (commented):
Looks good, just a few minor comments!

Also, in the last build, lint didn't pass. If that happens again, use `tools/lint.sh` and `tools/autofix.sh` to fix the issue. See https://neuropod.ai/docs/stable/developing/#lint-and-static-analysis for more details

source/python/neuropod/backends/torchscript/test/test_torchscript_dict_with_union_value_type_model.py
```diff
@@ -0,0 +1,83 @@
...
+        str_col = inputs['a']
+        return {'c': torch.cat([self.preprocess(str_col), num_col], dim=-1)}
+
+
+@unittest.skipIf(
+    LooseVersion(torch.__version__) < LooseVersion('1.10.0'),
+    "Torch supports UnionType after 1.10, skip the test if torch version < 1.10"
+)
+@requires_frameworks("torchscript")
+class TestTorchScriptDictWithUnionValueTypeModel(unittest.TestCase):
```
VivekPanyam, line 57: This test seems to fail on GPU w/ Torch 1.10.

The error is 
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument tensors in method wrapper___cat)
```

The inputs to `cat` in the model above should all be on the same device.

See the logs here for more details: https://buildkite.com/uberopensource/neuropod/builds/171#ac931bc6-ead8-4bff-9ba0-c6e8664f6d52/368-2477

voe09: @VivekPanyam could you grant me access to neuropod buildkite? Thanks!

VivekPanyam: Working on it

VivekPanyam (approved):
LGTM

